PR #34, opened 2015-12-15
Use subdir-objects as it will be the automake default in the future & add JPEG decoding

atupone:
The big change is due to a file shared between bzadmin and bzflag :(


atupone:
Rewrote. Moved bzflag and bzadmin executable to top directory

JMakey:
Both "make distclean" and "make maintainer-clean" fail for me when this change is applied.  Can you get them to work?

atupone:
They should work now.

blast007:
What is the reason behind this change?  Is it simply because BZAdmin currently includes files from src/bzflag/ in its build process?  If so, wouldn't it be better to just fix that?  Also, there's been discussion about the possibility of just removing BZAdmin for 2.6 if we can properly replace it.

atupone:
Yes, the problem is bzadmin, and also the tests directory
Maybe a symbolic link could be a fix for that without lot of changes, but I don't know if windows is supposed to support that.

macsforme:
Thanks for the contribution. However:

> @blast007 I've already fixed the subdir-objects issue in a way that doesn't create giant Makefiles at the top level directory
> @blast007 and JPEG support is pointless since I'm not adding JPEG support to the image site nor the Windows build

So we will not be merging in either of these changes at this time.

PR #37, opened 2016-01-10
Adds timestamps to chat/server messages

gumpu:
This change add timestamps (HH:MM:SS) to each chat or server message.
It is configurable (on/off) via the GUI Options menu.

So chat/server message will look like:

    23:18:59  Hello world


JMakey:
We want to use std::string variables instead of character arrays in new code.  Can you remove the need for the "hms" array?

gumpu:
Done.  And as a bonus it results in nicer looking code too.

blast007:
I ended up implementing this in a different way that allows the timestamps to be turned on and off retroactively.  This was implemented in 034af40.

PR #43, opened 2016-03-13
reproducible: make build follow SOURCE_DATE_EPOCH specification

a-detiste:
The SOURCE_DATE_EPOCH environment variable would
typically be defined in debian/rules or some other
distro equivalent rule file.

https://reproducible-builds.org/specs/source-date-epoch/

https://tests.reproducible-builds.org/rb-pkg/unstable/amd64/bzflag.html

@apoleon


a-detiste:
PS: during one build, the files named usr/​share/​doc/​bzflag-​server/​plugins/​*.txt disapeared, that's more worrysome. Fixing the easy 'Reproducible' bugs allows everyone to catch more complex bug like this next time.

JMakey:
Commit 3b24de2 makes man pages reproducible in a more consistent, if less automated, way.

a-detiste:
Thanks, that's even simpler !

blast007:
Commit bf8e847 implements SOURCE_DATE_EPOCH support, for both the game version string and the manual pages datestamp (at least on systems using autoconf, so, everything except Windows and OS X).

PR #47, opened 2016-04-07
Text file extensions

khonkhortisan:
Using .txt for text files keeps from cluttering the list of known extensions as if .WINDOWS was a filetype, and allows opening a file without having to be asked what program it should be opened with.


blast007:
Just drag the files into your text editor, use File > Open, or use an editor with a sidebar file browser (such as Komodo Edit).

khonkhortisan:
So, starting from windows explorer, navigate to where bzflag was cloned, notice a README, open notepad, File→Open, navigate to where bzflag was cloned again, then open the README.

allejo:
A lot of developers have an editor such as Notepad++ installed, where you can right click any text file and say "Open with Notepad++". If the naming scheme would be changed, that'd move away from GNU conventions

khonkhortisan:
Plugins already use a README format with file extension: README.CustomZoneSample.txt README.HoldTheFlag.txt README.Phoenix.txt. I left shell scripts alone, as there's no reason to run them on windows, and you can tell they're executables by font color (exec bit) and the first line. I left the irix platform alone, and doxyfile. One lowercase readme was capitalized. What was MANIFEST for?

blast007:
The plugin readme files are end-user documentation.  The files you are suggesting to change are developer documentation.  We will not be changing them.

khonkhortisan:
data/fonts/readme follows neither the plugin end-user documentation format README.PluginName.txt, nor the GNU convention developer documentation format README.PLATFORM.

PR #49, opened 2016-05-01
Visual Studio 2015

khonkhortisan:
Requires https://github.com/BZFlag-Dev/bzflag-dependencies/pull/2

src/c-ares/Makefile.msvc values taken from renamed msvc_ver.inc a788a1ac1caad4549aa76cd85fdc3d7f2f82b041
isnan: A boolean shall return from a function prefaced, "is".

Regex is now supported by C++11 and does not need to be a separate dependency, but that is outside the scope of this pull request.


JeffM2501:
The branch was updated to 2015 with a new set of dependencies, this request is no longer needed.

blast007:
This was taken care of with some recent commits.  So, 2.4.10 will be ready for Visual Studio 2015.

PR #53, opened 2016-06-23
-view three for greater or fewer than three monitors

khonkhortisan:
bzflag -view three currently puts the center view on the top-right, and the side views on the bottom corners. This requires one to have a number of monitors that is divisible into three parts, each part filling one of the views in virtual desktop space. For example, one large monitor for the center view, two narrow ones for the left view, and four small monitors for the right view. No monitor can cross a view boundary, because no views have the corresponding sides touching.
Assuming someone had a 2x2 monitor array for their desktop, to use the current -view three, they'd have to physically move the bottom-right monitor to the right, and the top-right monitor down into the new gap. Or, if they had a 3x1 monitor array for their desktop, they'd have to reassign them to quadrants instead of thirds.
I moved the three views so that they are next to each other in the correct order from the start. This would normally make the center view narrow enough that the gui freezes trying to render. To work around that, the gui now renders as if it is using -view normal. That is, the gui is rendered after the views, on top of them all, taking up the entire screen space.
This change makes -view three compatible with any size of rectangular monitor array, as long as they entirely cover the virtual desktop.
I believe this satisfies the original intention of multi-monitor widescreen gaming.
If anyone can find a reason why losing the original behavior would break someone's multi-monitor setup, please tell me.

### DIFF
--- a/src/bzflag/BackgroundRenderer.cxx
+++ b/src/bzflag/BackgroundRenderer.cxx
@@ -566,9 +566,7 @@ void BackgroundRenderer::renderSky(SceneRenderer& renderer, bool fullWindow,
 
     // draw ground -- first get the color (assume it's all green)
     GLfloat _groundColor = 0.1f + 0.15f * renderer.getSunColor()[1];
-    if (fullWindow && viewType == SceneRenderer::ThreeChannel)
-      glScissor(x, y, width, height >> 1);
-    else if (fullWindow && viewType == SceneRenderer::Stacked)
+    if (fullWindow && viewType == SceneRenderer::Stacked)
       glScissor(x, y, width, height >> 1);
 #ifndef USE_GL_STEREO
     else if (fullWindow && viewType == SceneRenderer::Stereo)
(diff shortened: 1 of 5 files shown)


khonkhortisan:
Ok, I found a reason myself. Currently, the GM lock and team flag arrow (and probably something else as well) tracks correctly until it stops at the edges of the first view. This change makes it track incorrectly horizontally over the whole fullscreen window. At least it goes over the other views now. The problem is that, even as a GUI element, it's still attached to an object in a scene, and should be part of a view. Or it could still be faked because the master UI knows where the views are.

macsforme:
Anyone have thoughts on what to do with this?

To me, `-view three` seems to have limited usefulness unless your monitors are already configured in a specific way. Eventually, I'd like to have menu support for configuring each monitor individually, where you can set the 3D view angle and whatever other info you want to display on each one. I'm not sure what to do with this in the meantime, though.

macsforme:
It sounds like our plan for `-view three` at this moment is simply to remove it, and look at other ways of implementing multiple monitor support in the future. As such, this PR will be closed for now. At the same time, thank you for your effort in identifying this issue and coming up with at least a partial fix. And we really need to be better at addressing pull requests in a timely manner...

PR #54, opened 2016-06-28
Add some visual enhancements

kaadmy:
Pretty much the name of this PR; add 3D grass.
Compiles and runs fine on Linux, will probably not compile on Windows, and will definitely fail on Mac. 

Whoever has access to Mac/Windows, please **attempt** to get this to work :)


kongr45gpen:
Works on windows for me

macsforme:
While I am impressed with what you were able to do within the confines of our current engine (which is very limiting), I do not believe that this patch is ready for "prime time" yet. Specifically, the appearance of the grass texture billboards is not appealing when you rotate your tank (they appear to be spinning in place... perhaps it would be better if they were smaller?), plus the fade in/out is very noticeable (the standard grass texture would probably have to be closer in appearance to the sprites for better blending). There should also probably be an effects option to control the density of the grass (or disable it) based on the capabilities of the user's machine. Additionally, unless I'm mistaken, there is no handling for the case where there is a custom ground texture in use (the grass should probably be disabled entirely in that case).

As far as the blending function change, I don't see where you've established a case for it other than saying it's an "improvement." I would want to see a little bit of the rationale for it and maybe some side-by-side comparisons; otherwise it seems somewhat arbitrary.

Feel free to keep your patch available and promote interest in it. If you want to continue improving it, we would be interested to see where it goes. If we ever do an engine replacement and have more capabilities, we would also love to have your help implementing some much better-looking grass that leverages those capabilities.

kaadmy:
Yeah, the highest options still lack a high grass density because I assumed my method of grass was inefficient so the grass option cannot go very high.

PR #56, opened 2016-10-03
Added the MPL 2.0 license alongside the LGPL license

macsforme:
Let me know how this looks, @timriker. I looked at other projects for examples of how dual licenses are specified, and it didn't seem like there were very many changes made in general, and certainly not at the level of every single source file.


timriker:
We should probably move COPYING to COPYING.LGPL and create a new COPYING that references both files. All the sources point to COPYING, and do not mention COPYING.MPL. If COPYING references both COPYING.MPL and COPYING.LGPL then I think we're ok.

Thoughts?

macsforme:
Good idea. Let me know how that looks, or tweak it to your liking.

JMakey:
I approve adding MPLv2.0 license terms to BZFlag source code as an optional alternative to the existing LGPLv2.1 license terms.

timriker:
Any other comments? I'm pretty happy with this. I'll leave this open for another week or so, and then merge it in.

macsforme:
Sorry for the force push, but after being lectured by @allejo about merge commit clutter ;-) and finding an error in one of my merges, I just did a rebase on the tip of 2.4 today so we'll get a cleaner history when this is merged in.

As we are coming up on one week since the last comment, I am not aware of anyone else who wants to weigh in on this, even though I've gone on IRC and pointed people in this direction several times.

macsforme:
@timriker, it's been two weeks and no further comments have been forthcoming. Furthermore, we have some other changes we want to commit that will create conflicts if this pull request is left open. If you have no further objections, could we get this merged in soon?

PR #58, opened 2016-10-17
-noSelfKills

khonkhortisan:
pull request copied from bzflag-import-1


macsforme:
Thank you, but this should go into master, not 2.4... can you fix that?

macsforme:
Never mind, I'll do it.

PR #59, opened 2016-10-30
Use cURL to (de/en)code strings

allejo:
Our current `TextUtils::url_encode()` has a buffer overflow issue when trying to encode non-ASCII characters and our `TextUtils::url_decode()` doesn't correctly decode non-ASCII charcters. Since cURL already provides [`curl_easy_escape()`](https://curl.haxx.se/libcurl/c/curl_easy_escape.html) and [`curl_easy_unescape`](https://curl.haxx.se/libcurl/c/curl_easy_unescape.html), I've changed our TextUtils functions to use those instead of our own implementations.

This issue directly affects `bz_urlEncode()` in our plug-in API and can lead to server crashes.

**bz_urlEncode() Input**

`This is a test. Here is a funny symbol: ß`

**Output**

`This+is+a+test%2E+Here+is+a+funny+symbol%3A+%FFFFFFC3%FFFFFF9F`

**Expected Output**

`This%20is%20a%20test.%20Here%20is%20a%20funny%20symbol%3A%20%C3%9F`

Can I get feedback on this as to whether this is the correct approach or not? Or how it can be improved?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,8 @@
 BZFlag 2.4.9
 ------------
 
+* TextUtils::url_encode and TextUtils::url_decode now use cURL's functions; this
+    change fixes a crash bug with bz_urlEncode() - Vladimir Jimenez
 * Add bz_reloadBadwords() to API - Vladimir Jimenez
 * Add ability to reload badwords list with '/reload badwords' - Vladimir Jimenez
 * Source code now dual licensed under the MPL 2.0 or LGPL 2.1 - Tim Riker
--- a/src/common/TextUtils.cxx
+++ b/src/common/TextUtils.cxx
@@ -17,6 +17,7 @@
 #include "TextUtils.h"
 
 // system headers
+#include <curl/curl.h>
 #include <string>
 #include <string.h>
 #include <algorithm>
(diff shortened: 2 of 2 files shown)


jwmelto:
Using a "+" for a space is common for url-form-encoded, but it's not a direct encoding.  I haven't looked at TextUtils::url_encode, but from your example it appears that there is a sign-extension problem when using wide characters. But it's probably not surprising, since we don't support wide characters systemically.

I'm a fan of using what is already written. The curl library functions should be expected to be idempotent (although I can't assert I've tested them). Looks like you've got a good unit test there.

jwmelto:
OK, you made me look.

Having to call curl_easy_init for every encoded string is a bit of overhead. Also, consider
http://stackoverflow.com/questions/14867404/curl-easy-escape-not-working-from-c-dll

That said, I dug around in the source code and it is acceptable to pass a NULL pointer for the first argument to curl_easy_(un)escape.

allejo:
Good call about passing NULL instead of using `curl_easy_init()`. I think internally, it's still creating a CURL object but I can't say for sure as I haven't dove into cURL's source. In your SO link, it mentions [this implementation](http://www.geekhideout.com/urlcode.shtml) which we could consider if we don't want to use cURL's functions. I'm not familiar with how DLLs work so not sure if this would affect us?

jwmelto:
@JeffM2501 would be better to speak to this, but in general you don't want to allocate memory on one side of "the DLL boundary" and use it on the other side. Also, you can't rely on instantiated templates (e.g., std::string) to have the same implementation on both sides. So if you allocate a std::string in the application and pass it to a DLL compiled with a different definition, it may "call" a function at a different address on your object resulting in unpredictable behavior.

Since you said this affects the API, and I'm pretty sure TextUtils is available for plugins to use, it may be better to just fix the bug in our implementation than to use curl. The implementation referred to by the SO page isn't so different from ours, so we were on the right track.

JeffM2501:
You can use curl in the API backend, just don't have it in the API header. Do all the curl work inside bzfs and make the API use the API string class, it's already solving the issue of std::string across the DLL boundry.

Or kill bzfs on windows and use whatever you want.

allejo:
Just for clarification, the API facing part of this issue returns a `const char*` but the internals of it and the TextUtils dependency uses `std::string`. The return types aren't changed in this pull request, it's just how the returned data is gotten that has changed.

JeffM2501:
Who allocates the data that const char* points to and who is supposed to de-allocate it? If you are expecting the plug-in to free it, you'll have a problem. If the memory behind the char* is fully owned by BZFS there is no issue.

allejo:
After talking with Jeff, this PR wouldn't affect DLLs since the API is returning a `const char*` via std::string's `c_str()`, so the std::string will clean up after itself. The std::strings that TextUtils and the cURL functions are returning/using aren't passing across the DLL boundary so we're not affected by that.

PR #61, opened 2017-01-16
[WIP] Add support for custom polls via plug-ins

allejo:
This pull request introduces the ability to create custom poll types to the API. A custom poll type is defined as adding your own action to `/poll <action>`. I've added a sample plug-in showing how to create a primitive `/poll mute` option.

## New API Events + Data Objects

- `bz_eAllowPollEvent` -> `bz_AllowPollEventData_V1`
- `bz_ePollStartEvent` -> `bz_PollStartEventData_V1`
- `bz_ePollVoteEvent` -> `bz_PollVoteEventData_V1`
- `bz_ePollVetoEvent` -> `bz_PollVetoEventData_V1`
- `bz_ePollEndEvent` -> `bz_PollEndEventData_V1`

## New API Functions

- `bool bz_registerCustomPollOption (const char* option, const char* parameters, bz_CustomPollOptionHandler *handler);`
- `bool bz_removeCustomPollOption (const char* option);`

## bz_CustomPollOptionHandler Usage

`bool PollOpen (bz_BasePlayerRecord *player, const char* action, const char* parameters)`

This function is called before every custom poll begins. When this function returns true, the server will proceed and begin the poll. If the plugin returns false, then the poll will not begin. This function is intended for allowing plug-in developers to check conditions or permissions before the poll is begun. Any messages or errors that occur should be sent to the player from this function, otherwise it will appear as if the poll failed to start.

* player - the player record of the player who would like to initiate a poll
* action   - the action is that being polled, e.g. 'mute'
* value    - the value or target of the action, e.g. 'talkative_player'

`void PollClose (const char* action, const char* parameters, bool success)`

This function is called after every custom poll has ended. This is where the business logic of a successful poll should be defined.

* action  - the action is that being polled, e.g. 'mute'
* value   - the value or target of the action, e.g. 'talkative_player'
* success - whether or not the poll succeeded

I'd appreciate feedback and help testing too ❤️


jwmelto:
Be sure to change the comments (in the header files, at a minimum) to reflect the change in API. For example, in VotingArbiter.h you changed "std::string getPollRequestor()" to "int getPollOrganizer" but the comments still say "return a string representing the player who requested the poll".

jwmelto:
this is a nit, but "const int" as a function parameter is silly. There is no valid reason to declare a pass-by-value parameter "const" since it's impossible to change the underlying value. It is an artifact of a global replace of "std::string&" to "int" in the signatures.

jwmelto:
This is one of my pet peeves, but if you are modifying a ctor, make every effort to initialize members in the member initializer list, and not in the function body.

include/bzfsAPI.h
```diff
@@ -275,6 +275,10 @@ typedef enum
   bz_eGameResumeEvent,
   bz_eGameStartEvent,
   bz_eGameEndEvent,
+  bz_ePollStartEvent,
+  bz_ePollVoteEvent,
+  bz_ePollVetoEvent,
+  bz_ePollEndEvent,
```
jwmelto: Doesn't this change require incrementing the API version? Previously compiled plugins are likely to be incompatible.

allejo:
@jwmelto I've made some commits (that I'll squash when this is ready to merge).

Regarding your comment about the API version, I've bumped `BZ_API_VERSION` but I didn't know that's what that number was for. Maybe @JeffM2501 could confirm?

As for initializing members, would I do all of the members or only the members that would have default values?

I've also fixed the comment in VotingArbiter and changed it to just an `int` instead of a `const int`

jwmelto:
sorry for the spam comments. I finally figured out how to group them in a review... but to member initialization: default initialization may be elided, but may be explicit for pedantic and documentation purposes.
http://www.cplusplus.com/forum/articles/17820/

allejo:
So taking `bz_PollStartEventData_V1` as an example, would I just initialize all those strings to empty strings and `BZ_SERVER` for the playerID?

src/game/VotingArbiter.cxx
```diff
@@ -58,16 +59,19 @@ bool VotingArbiter::forgetPoll(void)
...
+  _pollOrganizer = BZ_SERVER;
 
   /* poof */
   _suffraged.clear();
 
   return true;
 }
 
-bool VotingArbiter::poll(const std::string &player, const std::string &playerRequesting, std::string action, std::string playerIP)
+bool VotingArbiter::poll(const std::string &target, const int playerRequestingID, std::string action, std::string playerIP)
```
jwmelto: "const int" is superfluous. "int" is sufficient.

macsforme:
> I've made some commits (that I'll squash when this is ready to merge).

Kind of a side note, but I get the impression that you're planning on making a habit of squashing your pull requests when you merge them in. It would be one thing if all development was done in pull requests, and our main repository's history was basically a list of merges. However, since it's not, I personally think that commit history from within pull requests is just as useful as that of commits we push to the main repository, so I would encourage you to keep them intact.

include/bzfsAPI.h
```diff
@@ -275,6 +275,11 @@ typedef enum
   bz_eGameResumeEvent,
   bz_eGameStartEvent,
   bz_eGameEndEvent,
+  bz_eAllowPollEvent,
+  bz_ePollStartEvent,
+  bz_ePollVoteEvent,
+  bz_ePollVetoEvent,
+  bz_ePollEndEvent,
```
allejo: Move these declarations to the end of the enum

include/bzfsAPI.h
```diff
@@ -45,7 +45,7 @@
 
 class bz_Plugin;
 
-#define BZ_API_VERSION	26
+#define BZ_API_VERSION	27
```
allejo: after confirmation, no version bump is needed. it's only necessary when existing constructors, objects, or functions are modified

plugins/customPollOptionSample/customPollOptionSample.cpp
```diff
@@ -0,0 +1,80 @@
+// CustomZoneSample.cpp : Defines the entry point for the DLL application.
```
allejo: fix the copy paste

plugins/customPollOptionSample/customPollOptionSample.cpp
```diff
@@ -0,0 +1,80 @@
...
+  // This will introduce a '/poll mute callsign' option
+  bz_registerCustomPollOption("mute", "callsign", this);
+}
+
+void customPollOptionSample::Cleanup ()
+{
+  Flush();
+
+  // Remove the poll option when this plugin is loaded or else what other plugin would handle it?
+  bz_removeCustomMapObject("mute");
```
allejo: `bz_removeCustomPollOption`*

include/bzfsAPI.h
```diff
@@ -1538,6 +1611,21 @@ BZF_API bool bz_pollVeto( void );
...
+// custom polls
+
+class bz_CustomPollOptionHandler
+{
+public:
+  virtual ~bz_CustomPollOptionHandler() {};
+
+  // Should return false to prevent the poll from starting
+  virtual bool PollOpen  (int playerID, bz_ApiString action, bz_ApiString value) = 0;
+  virtual void PollClose (bz_ApiString action, bz_ApiString value, bool success) = 0;
```
allejo: since these values won't change/shouldn't be changed, making them `const char*` would work

include/bzfsAPI.h
```diff
@@ -843,6 +847,58 @@ class BZF_API bz_SlashCommandEventData_V1 : public bz_EventData
...
+  // 'kick', 'kill', 'ban', 'set', 'reset', or custom value from a plug-in
+  bz_ApiString pollAction;
+
+  // If it's a 'kick', 'kill' or 'ban', this will be the victim's callsign
+  // If it's a 'set', this will be e.g. '_mirror black'
+  // If it's a 'reset', this will be 'flags'
+  bz_ApiString pollTarget;
+
+  // This value will be set to the victim's IP if it's a 'kick' 'kill' or 'ban'
+  bz_ApiString pollValue;
```
allejo: `pollValue` since then has been removed since it's only available on certain occasions and having overload values are a pain, as you said. Besides, the information is only an IP so that can be fetched/stored by the plug-in

include/VotingArbiter.h
```diff
@@ -363,6 +366,15 @@ inline TimeKeeper VotingArbiter::getStartTime(void) const
   return _startTime;
 }
 
+struct PollOption {
+  std::string pollParameters;
+  bz_CustomPollOptionHandler* pollHandler;
+};
```
allejo: done

include/VotingArbiter.h
```diff
@@ -363,6 +366,15 @@ inline TimeKeeper VotingArbiter::getStartTime(void) const
...
+extern std::map<std::string, PollOption> customPollOptions;
```
allejo: Without making this extern, I wouldn't be able to access the map in other files as is. Do you have a better approach in mind for this?

allejo:
Made some updates with the given feedback. Still waiting on a few more opinions before this can be merged in.

allejo:
I think I'm all set. Any other feedback is welcome and would be appreciated before I merge this in the coming days.

JMakey:
I know this is tangential to this pull request, but this program:

```
#include <stdio.h>
static int f(const int value)
{ return printf("value = %d\n", ++value); }
int main()
{ return f(3); }
```

illustrates the type of situation where it may be reasonable to declare a function parameter as `const int`.  It indicates to the compiler that the `value` variable is not to be modified within the function, and gcc dutifully throws an error at the attempt to modify it.  Such a style can help prevent programmers from mistakenly thinking that the modified variable will be passed back outside of the function.

jwmelto:
So are you advocating for the style? Because if a programmer doesn't know the difference between "int value" and "int& value", it's unlikely that even the compiler error is going to help them.

The function signature is a contract between the caller and the writer of the method, not between the writer and himself. YMMV

JMakey:
A programmer can know the difference between those two declarations and still fail to be aware of which is in effect for a specific variable.  You may not need more safety in your programming style, but I appreciate it when I see code that shows an extra effort (a not-so-superfluous use of `const` in this case) to prevent errors both in the past and in the future.

PR #62, opened 2017-01-25
premake build system

macsforme:
This branch implements a cross-platform build system using [premake 5](https://premake.github.io), replacing the individual project files we maintain for Visual Studio, Xcode, and autotools. 

### Building

You will need to download the latest alpha build of premake 5 from [here](https://github.com/premake/premake-core/releases). macOS users will have to build the latest development version of premake 5 from source, because the two pull requests I submitted to them with fixes for Xcode have not yet made it into an actual release. Once you have obtained premake, make sure it's in your *PATH*. From the main BZFlag source directory, you can execute `premake5 --help` to see all of the available actions and configuration options.

#### Windows

First you will need to set up bzflag's dependencies like normal (build the bzflag-dependencies sources, and set up your BZ_DEPS environmental variable with the correct path). Next, change directory to the main BZFlag source directory, and execute `premake5 vs2010`. You can try to make project files for a different version of Visual Studio, but I believe our dependencies will only build with 2010 right now so that's probably the version you will have to use. In the build/ directory, open BZFlag.sln and then build as normal. Just like before, you can find the build product in bin_(Release|Debug)_Win32/.

#### macOS

You will need to install c-ares into /usr/local and SDL2.framework into /Library/Frameworks like you did before (see README.MacOSX). Next, change directory to the main bzflag source directory, and execute `premake5 xcode4`. In the build/ directory, open BZFlag.xcworkspace and then build as normal. You can find the build product in build/bin/(Release|Debug)/.

You can also follow the instructions for Linux below to build using GNU make on the command line on macOS.

#### Linux

The premake scripts lack the extensive library/system detection that was present in the old autotools scripts, so you will have to have your build environment set up correctly or you will simply get a compiler error (the reason should be fairly evident from the particular error you get). Follow the instructions in README.Linux, or if you already have the required packages set up to build with the old autotools scripts, you should be all set.

To build and install BZFlag using premake, use the following:

    premake5 gmake
    cd build
    make [-j# if you like]
    cd ..
    premake5 install

When using the gmake action, you can also specify the `--prefix=path` option to install somewhere else, and there is a `premake5 uninstall` action as well. See `premake5 --help` for more information.

### Benefits

- Ease of refactoring/modularization - there have been several initiatives to refactor or modularize our code, but having to keep three sets of project files in sync while manipulating files is a daunting prospect. With this system, files can be added, deleted, or moved and the changes will be automatically picked up when you re-run premake. You generally don't need to modify the project files at all.
- Version info in one place - the BZFlag version is now specified in only one place in the source code and project files: in the main premake script. Previously, it existed in some form in at least four places: configure.ac, include/version.h, bzflag.rc, and BZFlag.nsi. All changes to these other files are automated. This reduces the effort required to do a release.
- Consistent client/server string - this is a minor benefit, but the version string is now consistent across all platforms in the format Major.Minor.Revision.Date-Status-SystemBitness-CompilerVersion[-SDL[2]][-dbg]. This puts a little more useful information in the string, eliminates hard-coded compiler/IDE version information (it is all dynamically generated by premake), and allows for simpler parsing of the string on game servers or the list server for statistical purposes.

### Remaining Challenges

- premake5 is still in alpha status - while premake5 appears fully functional for our needs, it is still in alpha status. New alpha releases have been forthcoming every few months, and they seem to be attentive to moving it forward. However, changes to our premake scripts might be necessary if premake evolves in some incompatible way (this seems unlikely, or the changes required should be minor). Additionally, this means that premake5 may not be available in Linux distro package managers for a while.
- Linux package creation for BZFlag - I don't create the .deb or .rpm packages for the linux distributions, but it is possible that premake may not play well with the process for creating these (depending on whether they expect a traditional autotools-based system with its commands). We may need to evaluate how to move forward on this. On the other hand, package creation may become much easier if we can automate it with a custom premake5 action. If we eventually go with the container approach for Linux, the container assembly process could almost certainly be automated within the premake scripts.
- Less support for optional features - the previous automake system detected a vast array of installed libraries and system features for the build system, some of which we used to support optional program features (or at least we pretended to, since everything was pretty static on Windows and macOS). premake has less support for automatic detection of system features, and actually the idea is that you can run premake on one system and then build using those files on another. As such, we may need to depend on configuration flags passed to premake (several of which are already implemented) to enable/disable optional features. With some of the standardization we have been doing anyway (requiring a modern C++11 compiler, for one), and with a more limited set of platforms to support, I don't believe the extensive system feature detection provided by autotools will be critical for us.

### Next Steps

Beyond the scope of this branch, there are other things that should be tidied up.

- tools/ and misc/ - there are many additional tools and scripts included in our main source repository. Some of these could previously be built using the autotools system (if specifically requested), and some of them had project files for other IDEs. All of these need one of 1) to be added to the main premake system so they build with the project, 2) to get premake scripts of their own (should be simple for most of them), 3) need to be removed from our main repository and put in their own repositories if we have any future need for them, or 4) frankly, need to go away. The ones I see at a quick glance are:
  - bzwtransform
  - modeltool
  - TextTool2
  - 3ds2bzw
  - rrlog
  - many various scripts
- Documentation overhaul - our READMEs have a lot of information pertaining to the old build systems and platforms (even some long since unsupported) that is no longer relevant, and they need new instructions for building with premake. These files probably just need a full rewrite, keeping the pertinent information and including the new stuff, and writing them with consistent formatting and in a concise way to reduce the maintenance burden.
- Other platform support - Solaris and BSD support need to be looked at; if they worked under the old autotools system then support is probably not too far away (especially if we can just use SDL 1.2 or 2), but the premake scripts might need some platform-specific filters and definitions for those platforms. I think support for additional platforms beyond these should be dropped.

### Conclusion

This is a big change, so please try things out and let me know about any problems or concerns that arise. I am more than willing to address any details that need attention, no matter how minor.


JeffM2501:
Here we go, using VC2015 (VC10 is not a viable option on my system). these are the issues I found.

Readme doesn't show how to use premake, still has old build system.

The resulting project tree is kind of crap, it matches the folder structure and is not super useful. Most windows projects put the headers into a header filter for ease of use instead of doubling up a bunch of files.

The project has lost the ability to build just the client or just the server, the only thing it can do is similar to the old full build, not good for actual development when you want to debug both the client and server at the same time in different instances of VC loading different project sections.

The game sub project won't build, it needs an include file that looks to be in bzfs, "GameKeeper.h"

The windows dependencies don't include SDL2, but this build is setup to use it.

There is code that won't compile in VC12 or later and needs changes.

1) include <stdio.h> in win32.h and surround the #define snprintf	_snprintf with a #ifndef snprintf

2) anything that includes std::max or std::min needs to #include<algorithm> too in order to be standards compliant. LocalPlayer.cxx is a large offender.

3) calls to isnan in CommandsImplementation.cxx are ambiguous. It's best to make it std:: and include<math>

Could not test after that due to those issues.

JeffM2501:
Also add _WINSOCK_DEPRECATED_NO_WARNINGS to the net lib's preprocessor defines.

JeffM2501:
Did some hacky fixes to get past all the other issues, 

BZAdmin has a link issue
Severity	Code	Description	Project	File	Line	Suppression State
Error	LNK2019	unresolved external symbol ___iob_func referenced in function _Xinitscr	bzadmin	E:\Projects\bzflag_premake\build\pdcurses.lib(initscr.obj)	1	
Error	LNK2001	unresolved external symbol ___iob_func	bzadmin	E:\Projects\bzflag_premake\build\pdcurses.lib(pdcscrn.obj)	1	


I'd be really nice to NOT have to build bzadmin if one does not want to.

JeffM2501:
The VC10 windows dependency bins can't be used to build in VC14, at least in debug.
Blast has built a VC14 package that includes SDL2
http://static.bzexcess.com/bzflag-dependencies-2017-01-28-VC14.zip

With those dependencies, disabling bzadmin, and the aforementioned hacky fixes for the above issues, the system will build and run in VC2015.

macsforme:
To summarize the remaining issues on Windows/Visual Studio after some of them were resolved over IRC today:

- The project tree layout is not ideal; there is a folder called ".." (also present in Xcode), the source files and headers are together (they also are in Xcode and in autotools, but they always were), and some manually-created file categorization for bzfs source files is no longer there.
  - *My thoughts: the location() function in premake may help us straighten out the tree a little bit; as for the filters, if we're going to categorize source files within a project, why not just put them in subdirectories, and then all IDEs can benefit from them? Premake would pick them up automatically then, too. I can look into whether premake can create filters to keep source and header files separate, but that seems to me like a matter of personal preference, as I can just as easily see the benefit of having source files right next to their headers for easy switching back and forth and seeing where one might be missing.*

- The individual .sln files for the client, server, and bzadmin are not created by premake; these are useful for simultaneously debugging the client and the server, etc.
  - *My thoughts: these may be possible to create these using premake with the externalproject() or includeexternal() functions, or by leaving the existing .sln files in place and using the uuid() function for each project to keep each project's UUID static.*

- Visual Studio resources are not in standard locations.
  - *My thoughts: standard locations may be something we have to give up by using premake; however, we could put generated files (like bzflag.rc, which premake generates with the correct version number) in-tree if that would be easier to work with, but that would complicate the `premake5 clean` action.*

Please let me know if I missed any or if any other concerns come up.

allejo:
As a side note, `plugins/newplug.sh` is broken in this branch as it looks for `SAMPLE_PLUGIN/Makefile.am`

macsforme:
The following updates have been made:

- The branch was rebased onto the upstream 2.4 branch to get the changes for full Visual Studio 2015 compatibility (sorry, but you will probably have to do some resets and updates or clone it again).
- Individual solution files (.sln) for Visual Studio are now created like before for fullbuild, bzflag, bzfs, and bzadmin.
- The vpath() function was used to clean up the project layout within Visual Studio and Xcode (so there isn't a ".." folder any longer), and groups were set up similarly to the existing groups in the previous Visual Studio projects (the exception is the topical groups in the bzfs project, which I still argue should be either used consistently for all projects, eliminated, or at least the actual directory layout should reflect it).
- newplug.sh was updated so it works; the SAMPLE_PLUGIN README will be updated later along with the rest of the documentation.

As I went over the differences between the Visual Studio projects that premake creates and the projects that we already had, these were all of the differences I noticed:

- Plugin projects don't show a "reference" to bzfs.lib (although they do link - you can see it in their properties); I'm not sure why this is.
- All plugins link to plugin_utils, whether or not they actually require the functionality; this is a side-effect of the automated plugin detection and configuration, and seems harmless to me; however, I suppose that further logic could be implemented to parse source files for plugin_utils requirements and only link if necessary.
- All .cfg files in the plugin directories are shown in the IDE along with the .txt files, sources, and headers.
- A "Documentation" group was created under the bzflag project so that you can edit all of our documentation files from the IDE.
- The topical grouping of source files in the bzfs project was not re-implemented, so bzfs has "Source Files" and "Header Files" like the rest of the projects; I will continue to strongly argue against this inconsistency we had before (either all of the projects should do it, and really it should be reflected in the directory structure, or none of them should; it should be consistent).
- The "Header Files" groups in projects that had no header files were excluded; I'm not sure that you can create an empty group in premake.
- mediafile/AudioFile.* and WaveAudioFile.* are incorporated into the project, whereas they weren't before (anybody know why?).
- The platform project used to include files that were excluded from being built; while this is technically possible to do in premake, I didn't re-implement this; it's easy enough to see which files are excluded in the premake scripts, and other IDEs like Xcode never had this functionality; we can revisit this if we need to.
- SAMPLE_PLUGIN is built and included; I believe it always was in Xcode, and I'm not sure about autotools; this can easily be changed, but I would argue for consistency across platforms.
- In the old Visual Studio projects, individual headers from include/ were listed with their corresponding source files in the individual projects, but now each project shows its own directory's headers and also includes the include/ directory where the rest of them are; technically we could make it the way it was before, but I don't want to have a mix of automatic file detection and manual listing of files (except where it's absolutely necessary, like in platform); I would argue for moving header files out of include/ and back into the directories where their source files are like we had in 2.99, or else giving up on automatic file detection and listing files manually again so all IDEs are consistent (at the expense of having to track files manually and update the scripts due to the loss of automation).

Some things that still need to be done:

- @JeffM2501 suggested splitting the version information out of bzflag.rc and only auto-generating a version.rc so the rest could be updated by the developer; this sounds like a good way to go.
- Right now all generated files are put in the build/ directory, which makes it easy to clean up, but leaves some things in unintuitive locations; I would like to probably at least make another folder for project files (maybe within build/), and maybe leave other generated files (like the .rc and .plist files) next to their templates, which would just need a slightly more complicated `premake5 clean` action.
- It sounds like there may be some hesitation to having premake completely replace the rest of the projects at this time, so I may rework this to exist alongside our existing autotools scripts and Visual Studio projects for the time being (they can replace the Xcode project), and hopefully when premake 5 is easier to get we can ultimately transition over completely; unfortunately, we will lose out on a lot of the automation that premake provides (I'm thinking of version numbers scattered about that need to be updated at each release).

macsforme:
This pull request will be reworked so it replaces the Xcode project only, and coexists alongside the current gmake and Visual Studio project files, while remaining as an option for automatically creating project files for those latter two build tools.

PR #64, opened 2017-01-29
extend bz_ePlayerDieEventData to v2 

JeffM2501:
to add flag killed  with that @allejo wanted
will solve issue Flag dropped data in bz_ePlayerDieEvent #1 
is not breaking and rather trivial to implement.

### DIFF
--- a/include/bzfsAPI.h
+++ b/include/bzfsAPI.h
@@ -489,6 +489,15 @@ public:
    bz_PlayerUpdateState state;
 };
 
+class BZF_API bz_PlayerDieEventData_V2 : public bz_PlayerDieEventData_V1
+{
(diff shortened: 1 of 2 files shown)


allejo:
The reason I put the issue on hold until 2.6 is because clients send a drop flag message before they send the death message.

https://github.com/BZFlag-Dev/bzflag/blob/a7f17f522596dabc41f0c3ff6b427502eece1806/src/bzfs/bzfs.cxx#L3187-L3189

I'm not familiar with the client's code so I can't confirm but that's what bzfs makes note of and that's what I saw when I was testing things (unless I missed something). This PR as-is, doesn't work though

PR #65, opened 2017-01-29
Update Model tool and remove link to common

JeffM2501:
Update VC project to 2015,
force it to use the deps regex
remove the need for files from common
Moved text utils that is uses into it's own code.
this makes model tool independent of bzflag in case it gets moved to a tools repo.


JeffM2501:
Blast will be doing this better.

PR #68, opened 2017-02-01
Update VC14 projects with reasonable default filters (folders)

JeffM2501:
add filters for all the projects.
cleanup duplicate headers.
ignore Vc2015 db files


JeffM2501:
will submit another one with the missing files and no history of the badly added files

PR #71, opened 2017-02-11
split windows .rc files into version and application files

JeffM2501:
This gives all the exe projects a version resource so they have the correct version in details.
But only the client gets the icon as part of bzflag.rc


macsforme:
I don't see an actual version.rc included in your changes.

allejo:
closed in 27b7d70f031bbd506fff5947b5a611f2972c9b11

Issue #36, opened 2016-01-06
Bad Flags drop on Pause


allejo:
Bad flags should not be dropped when a player pauses

JeffM2501:
Fixed with PR #167 that is a reimplements #76 with the formatting and shared_ptr changes from master.

PR #75, opened 2017-03-26
Fix for Issue #36

henfredemars:
It turns out that BZFlag dropping flags while paused is broken. It never worked correctly because paused players do not send updates, so the server drops our flag for not responding after five seconds regardless of BZDB_PAUSEDROPTIME. Further, the client didn't check if the flag was bad when initiating a drop (only happens when BZDB_PAUSEDROPTIME < 5 seconds). 

This pull fixes _pauseDropTime by not counting paused players as not responding for failing to send updates within the five-second window. 

This pull modifies the client not to send flag drops for sticky flags while paused. 

As a result, a sticky flag can be held indefinitely by paused player. You might want to consider if a paused player should send updates, and if players can be simultaneously paused and not responding. The current implementation (before this change) does not distinguished pause responsive from paused not responding.

### DIFF
--- a/src/bzflag/LocalPlayer.cxx
+++ b/src/bzflag/LocalPlayer.cxx
@@ -98,8 +98,6 @@ void			LocalPlayer::doUpdate(float dt)
 {
   const bool hadShotActive = anyShotActive;
   const int numShots = World::getWorld()->getMaxShots();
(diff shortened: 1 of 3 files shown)


henfredemars:
Ah, it turns out the server will still take the flag when unpausing. That should be fixed too.

PR #76, opened 2017-03-26
Fix flags dropping while paused

henfredemars:
It turns out this bug was deeper than I expected :smiley: I am opening a new pull request. 

BZFlag server conflates idle time with the last message update time that is used to determine if a player is not responding. Because of this, paused players were always considered not responding by server flag logic, and were forced to drop the flag after five seconds regardless of BZDB_PAUSEDROPTIME and even if the client held a sticky flag! Further, _pauseDropTime was broken and was not used by the server if it exceeds the non-responding time (five seconds). A player holding a bad flag could pause to have the flag taken after a few seconds, even if BZDB_PAUSEDROPTIME is set to a large value, and even if bad flag shaking is not enabled on the server!

I have split 'lastupdate' into two timers: one to track idle times, and another to track player updates. Paused players are considered idle but no longer get marked as not responding (by flag logic) for being paused. The effect is that _pauseDropTime now works for times longer than five seconds, and players cannot use pausing to shake a bad flag early.

This pull fixes issue #36 and fixes _pauseDropTime which never worked properly before.


JeffM2501:
I've reimplmented these changes as PR #167 because there were many changes in the 2.6 branch since this was made

PR #84, opened 2017-09-03
Implement libpng for reading textures and writing screenshots

blast007:
This replaces our custom png reader and makes use of libpng instead.  Currently some of our textures trigger a libpng warning to stderr.  We should fix those textures, and possibly create a custom function for handling the libpng warnings.


macsforme:
Would you please explain how this is an improvement over what we already have?

I may very well agree with you in the end, but since you are proposing that we replace working code with an external library (which incurs the cost of managing an additional external dependency in all of our build systems), I feel like the justification should be clearly stated so the costs/benefits can be clearly understood.

atupone:
Probably the standard libpng can handle a lot more variants that our code currently handle.
If that is true you don't need to rework the images before feeding those to bzflag.
An other reason could be that exploit against our current implementation could be fixed by a more standard/used code

blast007:
Better error handling, knows about more edge cases, well tested library that gets a lot more eyes looking at it for issues.  We won't have to dive into PNG specs and implement new features ourselves.  Texture reading speed is pretty much identical.  Writing screenshots with the new code is 2 to 4 times faster, while keeping a similar filesize (a 5% filesize increase in one test with the new code).

macsforme:
Thank you. I think libPNG has good support on all our target platforms, so hopefully the maintenance cost will not be significant. I just think it is good to lay out the benefits clearly before adopting an additional dependency so we have a good idea whether they outweigh the costs. I want to avoid repeating choices like adding GLEW as a dependency, where we only ended up using it to test for the presence of a single extension (which had a low benefit/cost ratio).

PR #86, opened 2017-09-10
Fix compilation on OS X when using autotools / gcc

mikeage:


mikeage:
Testing:
```
./autogen.sh
./configure
make
```
Pre-requisites: installation of c-ares and sdl2 from homebrew

macsforme:
Would you explain why you want to link to GLUT? Also, will this change respect the SDL 1.2 versus SDL 2.0 build setting? It appears that you have forced SDLMain.m to build regardless of which one is selected.

mikeage:
GLUT: I thought I'd seem link errors without this (glu* symbols), but indeed, you are correct, and it's not needed. I'll push a new commit in a few minutes. 

SDL/SDL2: from what I saw, the XCode build didn't support SDL, only SDL2, and I was copying from there (my motivation for this PR was a pure command line build; it didn't seem possible to build using `xcodebuild` since there were no schemes saved and the targets didn't include dependencies)

mikeage:
Ok, fixed. Sorry about that -- it's been 15 years or so since I thought about anything related to OpenGL, and I got glu and GLUT mixed up! (I also put in a proper test for gl.h for darwin, rather than hardcoding it, which is of course better)

src/bzfs/Makefile.am
```diff
@@ -17,7 +17,12 @@ else
 plugin_files =
 endif
 
+if APPLE
+bzfs_LDFLAGS = -export-dynamic -static -framework CoreFoundation -framework Carbon
```
mikeage: I actually really don't like this; let me replace this with entries in `configure.ac` (ditto for the other two Makefiles)

macsforme:
Xcode is set up to build with SDL 2, but it's also set up to be able to switch to SDL 1.2 with a few build setting changes. SDLMain.m is included in the Xcode project, but isn't built by default because it's an SDL 1.2 file. It doesn't look like your makefile changes will honor the --with-SDL={1,2} setting. Can you either make that work, or else force --with-SDL=2 on macOS? It seems that there might be some unexpected behavior otherwise if people try to pass --with-SDL=1 to configure. Otherwise, it looks pretty good to me.

If it helps at all, here's a gist with the start of a patch I was working on to try to do what you are doing (I can't remember the particular error I was running into when I gave up): https://gist.github.com/macsforme/08297cd4bb2d099910b508959249fdc2

mikeage:
Gotcha. I'll work on that today.

mikeage:
Ok, ready. Thanks for pushing me to do this :)

I tested with the following builds:
```bash
 ./autogen.sh && ./configure --with-SDL=1 && make
 ./autogen.sh && ./configure --with-SDL=2 && make
 ./autogen.sh && ./configure && make
```
And for each, ran a sanity test by loading `src/bzflag/bzflag`, and checking that `` showed it linked to the proper SDL framework.

macsforme:
Also tested here and seems to work fine. Thanks for your good work.

One last question I can think of: I assume this requires SDL to be installed as a library rather than as a framework? I don't think there's a framework check in our code currently (there used to be), but regardless, I wasn't able to get it to find my SDL framework when I played around with various `-F` options.

Checking for an SDL framework may be beyond the scope of your project, so it's likely acceptable as-is, but that may need to be a feature request for the future if we are going to support command-line builds on macOS going forward.

mikeage:
True; personally, I do almost everything on my mac using homebrew which installs the libraries, but I know that the more traditional way of developing is with frameworks (though I would suspect that the majority of the framework folks would use the xcode build, regardless (which doesn't support the libraries)). It doesn't seem like autotools has a nice way to check for frameworks, from what I can  find; do we want the same duplication we have by the checks for opengl (there, it's unavoidable...)

mikeage:
Thanks!

PR #89, opened 2017-10-09
Add Travis to test builds on macOS + Linux

allejo:
Regarding earlier discussions, adding simple Travis tests to make sure things continue to build after commits on both Linux and macOS.

As an example: https://travis-ci.org/allejo/bzflag/builds/285422595

Currently, can't run `./configure` with the `--enable-debug` flag since it'll treat warnings as errors; this affects both Linux and macOS, see here:

- https://travis-ci.org/allejo/bzflag/jobs/285413694#L1687

Additionally, should we add IRC notifications? If so, how's BZnotify configured/what endpoint should we point to?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,5 +9,8 @@ indent_style = space
 indent_size = 2
 tab_width = 8
 
-[*.md,*.markdown]
+[*.{md,markdown}]
 trim_trailing_whitespace = false
+
+[*.yml]
+indent_style = space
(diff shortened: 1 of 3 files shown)


allejo:
For Linux, only Trusty and Precise are available so that's what we're stuck with

allejo:
Closing this. By the time we start working on 2.6, Travis OS options will have changed and so will have our build deps.

PR #91, opened 2017-10-09
upgrade to VC2017

JeffM2501:
Now with less history


blast007:
The project files are Visual C++ 2017 as of f6defa4.

PR #110, opened 2018-05-07
Use GLEW to get openGL entry points.

atupone:
I modify bzflag in order to get openGL entry points, and available extensions using GLEW.
Needs to fix windows file


atupone:
Would like to merge in the 2. 4 branch, but needs windows dependencies fix

PR #148, opened 2018-06-01
Use of Vertex Buffer Objects

atupone:
When sending vertex to openGL use only VBO.
To simplify the work I'm using:
- only what was used with Quality High and Experimental
- blending
- texture
- Radar Style enhanced view
List has been removed as glBegin/glEnd paradigm
This pull request is based heavily on glm usage (introduced as another pull request)
The missile mathematical rework is another pull request on which this is based, but can be easily removed


macsforme:
On consensus, this PR will be closed and we will wait for the OpenGL 2.0 branch (which incorporates these changes) to be ready, and evaluate that for merging into the master branch.

PR #149, opened 2018-06-02
In bzfs.6, change low-level tab hackery into TBL markup...

eric-s-raymond:
...so the man page can be lifted clean to DocBook-XML and produce
better HTML. This will also avoid problems with third-party
manual-page viewers such as xman and tkman.


JeffM2501:
2.4.x is not getting any more planned updates. Current work is being done in master branch for a 2.6 release, so we'll apply this to that branch

PR #154, opened 2018-06-11
Extended shot api

JeffM2501:
adds extended shot API (potion, velocity, expire etc..)
fixes bugs related to having clients create non-slot shots.


JeffM2501:
Nope.. nope nope.. wrong branch.

PR #158, opened 2018-06-15
Use OpenAL (and ALURE) for audio

blast007:
This still needs some work (for instance, making sure the audio falloff rate is close enough to the old audio code, possibly zeroing the velocity of some sounds).  I also encountered an issue with shot sounds from a remote client repeatedly playing, but I'm not yet sure if that's related to the OpenAL change.

It also currently lacks a sound priority system.


blast007:
Hmm, I'm also getting a segfault when I try to select a server in the server list.

blast007:
The segfault when selecting a server has been fixed.

blast007:
I'll possibly resume working on this at a later time.  One big difference with OpenAL and our audio code is that it doesn't take into account the speed of sound for propagation.  With OpenAL, if you tell it to play a sound, it will start playing immediately even if the source is on the opposite side of the map.  Our current code takes that into account and will delay the playback based on the distance from the source.

PR #170, opened 2018-07-27
Add shot limit on HUD

The-Noah:
Add the number of shots left for the current flag on the HUD. Remove the messages saying the number of shots left.


allejo:
Since this affects the client visually, please share screenshots of what your changes do. This way we get an idea of how it'll look and be able to give you feedback all without having to download your patch and test it ourselves.

The-Noah-BZ:
bzfi0019

JeffM2501:
It needs some kind of label or icon next to the number showing that it's a limit, not just a random floating number.

allejo:
I'm personally hesitant of the location of this label. Thoughts on placement? I'd favor a place where other flag information is already stored, e.g.

- Bottom of the mouse box, above where flag help messages would appear
- Upper right next to the flag name

The-Noah-BZ:
Hmm. I could add a simple option to switch it's location between where you said, that way if the player doesn't like it in one spot he/she could move it to the other. What do you think?

JeffM2501:
BZFlag does not need more options.
Show us screenshots or mockups with both locations.
I prefer lower near the shot reloads, since that's already shot related.

But there may be more room near the flag

allejo:
Since this PR has grown stale and it's something we're going to need to discuss further with some more mockups, I'm placing this one on hold.

TheGiraffe3:
So what's the status on this?

PR #178, opened 2018-09-09
Use AC_CHECK_HEADER and AC_SEARCH_LIBS for finding GL/GLU/GLEW and don't depend on pkg-config

blast007:
I replaced some of our AC_LINK_IFELSE macros with AC_CHECK_HEADER and AC_SEARCH_LIBS.  I currently special-cased macOS to look at the OpenGL/ include directory instead of GL/, but perhaps that could be simplified by using AC_CHECK_HEADERS instead of AC_CHECK_HEADER.  I can't test on mac to know if it even works in the current state.  I didn't want to depend on pkg-config which isn't included on a basic Debian install nor needed for a server-only build and without it autoconf would error out. This also reverts a change that made a curses library required even if you didn't care about bzadmin, going back to the old method where it would just build bzadmin without the curses UI.

I did also bump the minimum autoconf version to 2.69 (which is the latest version of autoconf, from April 2012).  I'm using a feature first mentioned in the documentation of autoconf 2.65 where AC_SEARCH_LIBS defines a ac_cv_search\__function_ variable, so we should be able to drop back to that if 2.69 is a problem.

This also shows more helpful information in the summary about what dependency is missing (GL, GLU, and/or GLEW).  It will still checks for game client dependencies and shows which are missing even if --disable-client is passed, but I didn't want to go crazy with cleanup/refactoring for this.

I have tested this on Debian Stretch (both a Gnome install and a minimal install), Debian Jessie, Ubuntu 14.04 Desktop, CentOS 7, and Fedora 24.  I did test it on Ubuntu 12.04 Server with the required autoconf version dropped back to 2.68, and it worked, but 12.04 hasn't been supported for over a year.  I'm hoping someone can test a *nix style build on macOS to make sure it works there.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,8 @@ BZFlag 2.4.15
 
 * Revert a80ab87 (truncation warning fix) as that generated broken replay file
     headers. Detect and work around that bad header. - Scott Wichser
+* Use pkg-config, if available, to detect ncurses
+    - Scott Wichser, Joshua Bodine, Alfredo Tupone
 
 
 BZFlag 2.4.14  "Hello Motto" (2018-05-02)
--- a/autogen.sh
+++ b/autogen.sh
@@ -75,7 +75,7 @@
 
 # set to minimum acceptible version of autoconf
 if [ "x$AUTOCONF_VERSION" = "x" ] ; then
-    AUTOCONF_VERSION=2.58
+    AUTOCONF_VERSION=2.69
 fi
 # set to minimum acceptible version of automake
 if [ "x$AUTOMAKE_VERSION" = "x" ] ; then
--- a/autogen.sh
+++ b/autogen.sh
@@ -75,7 +75,7 @@
 
 # set to minimum acceptible version of autoconf
 if [ "x$AUTOCONF_VERSION" = "x" ] ; then
-    AUTOCONF_VERSION=2.69
+    AUTOCONF_VERSION=2.68
 fi
 # set to minimum acceptible version of automake
 if [ "x$AUTOMAKE_VERSION" = "x" ] ; then
--- a/configure.ac
+++ b/configure.ac
@@ -125,11 +125,7 @@ dnl ***********************
 BZ_CONFIGURE_STAGE([arguments], [1 of 9])
 
 # provide a with-curses option, test for curses
-PKG_CHECK_MODULES(ncurses, ncurses,
-  [CURSES_LIB="$ncurses_LIBS"]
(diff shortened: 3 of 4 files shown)


atupone:
if glew is required why there is a need for checking GL ? glew should be enough

blast007:
@macsforme, I had the $GLIBS in the wrong spot, so I'm curious if it works now on macOS.

@atupone, checking for GL/GLU makes sure they also exist and are usable.  That might be unnecessary on macOS, Debian, or Gentoo.  I did notice that the binary glew package/lib didn't depend on GLU on Debian, but the dev package has a dependency on GLU.  But consider a distribution like Slackware that doesn't do dependency tracking by default.  GLEW does have an option to build without GLU as well.  Also, without using AC_SEARCH_LIBS for GL/GLU, I end up getting this on my Debian system:
```
/usr/bin/ld: SceneRenderer.o: undefined reference to symbol 'glColorMask'
//usr/lib/x86_64-linux-gnu/libGL.so.1: error adding symbols: DSO missing from command line
```

As for the ncurses issue you mentioned on IRC:
```
< tupone> blast007: for curses we should use the pkg-config, the curses.m4, if needed, should be revisited (IIRC)
< tupone> the previous method did not detect the need of an extra library from ncurses
< tupone> something like libtinfo.so
< blast007> what platform is that an issue with?
< tupone> I remember a bug issued on gentoo by someone.
< tupone> https://bugs.gentoo.org/632022 I think this was
```
I have the curses.m4 now attempt to use pkg-config if available, falling back to the old way if it doesn't find it or if pkg-config isn't available.

macsforme:
That works better. OpenGL/gl.h and OpenGL/glu.h are now located properly. If OpenGL/glew.h is changed to GL/glew.h, that header is also found in the default install location. I think that should be changed to GL/glew.h, unless someone has any insight whether the 3rd-party macOS package managers (like homebrew) put it under OpenGL/glew.h (maybe @allejo can answer?).

The GLEW library is also found properly, but the library searches for the glBegin() and gluScaleImage() functions fail. I read the code and I think I understand the intention. Looking at GLEW's successful pass in my config.log, the first run is:

    configure:21406: g++ -o conftest -g -O2 -std=c++0x   conftest.cpp -lm  >&5

And then the second run is:

    configure:21406: g++ -o conftest -g -O2 -std=c++0x   conftest.cpp -lGLEW -framework OpenGL  -lm  >&5

So it looks like AC_SEARCH_LIBS() isn't applying `other-libraries` to the first (and only, in the case of GL and GLU) run. Would there be a different way we can check the presence of those functions with a given set of GLIBS?

PR #187, opened 2019-01-05
Add new bz_eKillReassignmentEvent to the API

allejo:
Can I get a second opinion on this and its benefits on adding this notification event to the API?

See the discussion in https://forums.bzflag.org/viewtopic.php?f=78&p=176539

/cc @Zehra

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,7 @@
 BZFlag 2.4.19
 -------------
 
+* Add new bz_ePlayerDeathFinalizedEvent API event - Vladimir Jimenez
 * Output correct error when loading a nonexistent bzfs plugin - Vladimir Jimenez
 * Remove obsolete bzfs option from Start Server menu on Windows - Scott Wichser
 
--- a/include/bzfsAPI.h
+++ b/include/bzfsAPI.h
@@ -337,6 +337,7 @@ typedef enum
     bz_eServerShotFiredEvent,
     bz_ePermissionModificationEvent,
     bz_eAllowServerShotFiredEvent,
+    bz_ePlayerDeathFinalizedEvent,
     bz_eLastEvent    //this is never used as an event, just show it's the last one
 } bz_eEventType;
 
--- a/src/bzfs/bzfs.cxx
+++ b/src/bzfs/bzfs.cxx
@@ -3379,6 +3379,9 @@ void playerKilled(int victimIndex, int killerIndex, int reason,
         killer = realPlayer(killerIndex) ? &killerData->player : 0;
     }
 
+    // Call a new event when all the plugins have finished reassigning any kills
+    worldEventManager.callEvents(bz_ePlayerDeathFinalizedEvent, &dieEvent);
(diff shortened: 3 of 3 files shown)


JeffM2501:
Your API concept is flawed. You have created an API that requires all plugins that want to track kills to understand the internals of bzfs and track 2 events. The normal event, and the modified event. This puts extra work on every plug-in. The goals of an API should be to make plugins be as unaware of the internals as possible and for them to do less work, not more.

A better design is to have a single "finalKilledEvent", or "KillScoreChange" event that is called with the effective killer after the current die event is triggered (and possibly modifies the data). You would call this for every death, not just modified ones. That way the new event just replaces the existing die event, and tracks all kills.

Then plugins only have to check one event, the final notification event. In this way the core of BZFS can change with out affecting plugins.

If a plugin wants to know if a kill was modified, it can compare the 2 events, or leave the origonal and effective killers in the new notification event.

allejo:
@JeffM2501 I like your idea way more than my current implementation. With a "final kill event", should it contain the same data as the regular `bz_ePlayerDieEvent` event in addition to the kill reassignment data so it could be a drop-in replacement for `bz_ePlayerDieEvent` unless a plug-in explicitly wants to modify the killer data?

JeffM2501:
It depends if plug-ins ever need to care that a kill was reassigned. I don't think they would care, so really there is no need for extra data. You just need a single event that is after any processing from other events, with the final killer and victim data. You may not even need a new event, you already have a kill event. Make a v2 version of that data that has all the die event data, and then just ensure that it is triggered after all death processing.. That makes "die" be the event from a player that they died, and "kill" is the event when someone is killed. Most deaths would be triggered by a "die" message, but you could have other events that kill with out a die message, (server commands, etc...).

allejo:
I was following along until I got to this point.

> You may not even need a new event, you already have a kill event. Make a v2 version of that data that has all the die event data, and then just ensure that it is triggered after all death processing.. That makes "die" be the event from a player that they died, and "kill" is the event when someone is killed. Most deaths would be triggered by a "die" message, but you could have other events that kill with out a die message, (server commands, etc...).

When you mention the kill event, are you referring to `bz_eKillEvent` which is **only** called when `/kill` is used? Are you saying to call the `bz_eKillEvent` after the `bz_ePlayerDieEvent`s have finished processing? Because I'd disagree with using an existing event for this purpose because it could potentially confuse existing plugins, no?

JeffM2501:
Sure, if that will screw up isotherm the plugins then use a new event, whatever floats your boat.

allejo:
How does this sound? Fire off a new `bz_ePlayerDeathFinalizedEvent` after every `bz_ePlayerDieEvent` with the same data that `bz_ePlayerDieEvent` received (and possibly modified).

allejo:
a gentle nudge for this PR, @JeffM2501

JeffM2501:
It’s your project, you don’t need my approval.

PR #191, opened 2019-03-18
Reworked premake build system

macsforme:
This is a rework of #62. This premake5 system replaces the Xcode project only, and coexists alongside our Visual Studio projects and our autotools/gnu make scripts. However, it does support generation of project files for those latter two systems as well (for those who want to experiment with it), although they will be located in a separate area so they will not conflict with what we have.

The [README.premake5](https://github.com/macsforme/bzflag/blob/premake_third_try/README.premake5) file has more information.

The only other step that will be required will be to adjust the path for the Windows dependencies directory so it conforms to `<bzflag source>/dependencies/output-<system>-<configuration>-<architecture>`, as we discussed on IRC a while ago. This may also be a good time to adjust the Visual Studio project files so they look in the root bzflag source directory for the dependencies instead of relying on the `BZ_DEPS` environmental variable, as we also discussed a while ago.


macsforme:
Squashed and merged in 2553514a.

PR #193, opened 2019-04-12
Improve default settings for a better first-time experience

blast007:
This PR makes the following adjustments:
- Radar Style from Enhanced to Fast Sorted
- Radar shot length from 0 to 5
- Radar shot size from 0 to 4
- Larger radar size
- Leading shot line
- Cache size from 32 to 512 MB
- Removes the ineffective "performance test" at first client run and just assumes a halfway modern system with OpenGL acceleration (which leaves blending and smoothing on, dithering off, shot length of 6, and shadows and lighting on their best setting)


src/bzflag/defaultBZDB.cxx
```diff
@@ -34,15 +34,15 @@ DefaultDBItem defaultDBItems[] =
     { "volume",           "10",           true,   StateDatabase::ReadWrite,   NULL },
     { "latitude",         "37.5",         true,   StateDatabase::ReadWrite,   NULL },
     { "longitude",        "122",          true,   StateDatabase::ReadWrite,   NULL },
-    { "radarStyle",       "3",            true,   StateDatabase::ReadWrite,   NULL },
+    { "radarStyle",       "2",            true,   StateDatabase::ReadWrite,   NULL },
```
allejo: Are the colors used in our "fast sorted" radar style standardized amongst other games? Otherwise, I'd find it confusing for new players to understand what all of the colors mean.

blast007: From my experience, most games do not having a concept of depth in their mini-maps, short of maybe an indication that another entity is above/below the player, or maybe some sort of 3D-ish radar in a space battle game.  Most games have some static top-view representation of the world and overlay icons on top of that to indicate player/object/waypoint locations.

allejo: Hmm then my personal preference for a default radar style would "Enhanced" since it shows depth just with different alpha values instead of introducing greens, reds, purples, etc.

blast007: I agree.  I also just checked and death physics are only indicated on Enhanced radar.  I'll make the change shortly.

The-Noah:
Looks good. Personally I think the `radarsize` should be a bit bigger then you set, perhaps `10`.

PR #194, opened 2019-04-17
Use glm instead of our implementation (vectors.h)

atupone:
glm seems vastly used to operate with opengl.
This pull request is a step to use it instead of our own implementation


atupone:
Applied to master branch

PR #195, opened 2019-04-29
Make Visual Studio look for dependencies at source root rather than at $BZ_DEPS

macsforme:
Per our IRC discussion, this eliminates the need to set the BZ_DEPS environmental variable with the location of our dependencies on Windows, and instead the Visual Studio projects look at our source root for them, e.g. <bzflag source>/dependencies/output-<system>-<configuration>-<architecture>. This matches what the macOS script does, and it also allows us to include dependencies for multiple platforms in one directory (for example, a desktop OS and a mobile OS being tested on the same machine).


macsforme:
Corresponds to https://github.com/BZFlag-Dev/bzflag-dependencies/pull/8.

PR #196, opened 2019-05-06
Make radar size and console height be controlled by BZDB variables as discussed on IRC

The-Noah:
Previously the radar size and panel height were set initially by their BZDB variables at startup only and then changed directly from the settings menu. Now the BZDB variables are modified from settings instead and use a callback to change the SceneRenderer radar size and panel height.

### DIFF
--- a/src/bzflag/SceneRenderer.cxx
+++ b/src/bzflag/SceneRenderer.cxx
@@ -323,9 +323,9 @@ void SceneRenderer::setRebuildTanks()
 
 void SceneRenderer::bzdbCallback(const std::string& name, void *)
 {
-    if(name == "radarsize")
+    if (name == "radarsize")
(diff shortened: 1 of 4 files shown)


src/bzflag/SceneRenderer.cxx
```diff
@@ -317,6 +321,15 @@ void SceneRenderer::setRebuildTanks()
 }
 
 
+void SceneRenderer::bzdbCallback(const std::string& name, void *)
+{
+    if(name == "radarsize")
```
allejo: There should be a space between the `if` and open parenthesis

The-Noah: Sorry about that.

allejo:
Thanks @The-Noah, I've applied the patch manually so this can go into the 2.4 branch instead of master.

PR #197, opened 2019-05-07
Use markdown for information and README files

The-Noah:
Use markdown instead of plain text files.


allejo:
Thanks for your work on this! However, we're going to put this PR on hold for a bit while we figure some things out first.

- Not all of the formatting and whitespace we have in our plain text files translates well to markdown
- We want to revisit some of the content on these files. What can we get rid of or what should we rewrite.

macsforme:
I'm open to having our documentation files in markdown format, but I don't think it should be a straight-across conversion from our existing files. I would rather see the files be rewritten with the markdown format in mind, while at the same time trimming down the amount of information so it's more concise. Determining what should go into these files is something that will need to have input from the core developers, although I think the actual writing could possibly be done by other contributors once the general outline and content is determined.

allejo:
Closing this PR as we'll be revisiting our READMEs and not currently looking for a 1:1 rewrite.

PR #200, opened 2019-06-24
Add Slovak translation

jose1711:


data/l10n/bzflag_sk.po
```diff
@@ -0,0 +1,3205 @@
+# Slovak translation of BZFlag
+
+msgid "xx"
+msgstr "Zaklad prekladu"
```
blast007: Instead of:
```po
msgid "xx"
msgstr "Zaklad prekladu"
```

This could be:
```po
msgid "sk"
msgstr "Slovak"
```

That way, in the language selection menu, it will show "Slovak" instead of "sk".

data/l10n/bzflag_sk.po
```diff
@@ -0,0 +1,3205 @@
...
+msgstr "Power"
+
+msgid "Euro"
+msgstr "Euro"
+
+msgid "Undo"
+msgstr "Undo"
+
+msgid "Left Mouse"
+msgstr "Lave tlacidlo mys"
```
blast007: I noticed that for "Left Mouse", the translation used "mys" while "Middle Mouse" and "Right Mouse" used "mysi".  I'm unsure if that's a typo.

data/l10n/bzflag_sk.po
jose1711: thank you, fixed

data/l10n/bzflag_sk.po
jose1711: great observation - yes, there is (or was) a typo. now fixed. thanks!

blast007:
Thanks!  The pull request has been merged.

PR #208, opened 2019-07-28
Cleanup for consistency with mesh rendering

Zehra:
Fixes inconsistent rendering in OO with mesh and non-mesh objects.
In reference to the [Eighth dimension cleanup/removal](https://forums.bzflag.org/viewtopic.php?f=3&t=20170) forum thread.

For the next major release some cleanup will be needed, but for now this is a minor edit which works well as it does not introduce any game play inconsistencies which might result otherwise.

### DIFF
--- a/src/bzflag/playing.cxx
+++ b/src/bzflag/playing.cxx
@@ -6075,17 +6075,6 @@ void drawFrame(const float dt)
 
             // add explosions
             addExplosions(scene);
-
-            // if inside a building, add some eighth dimension scene nodes.
-            const std::vector<const Obstacle*>& list = myTank->getInsideBuildings();
-            for (unsigned int n = 0; n < list.size(); n++)
-            {
-                const Obstacle* obs = list[n];
-                const int nodeCount = obs->getInsideSceneNodeCount();
-                SceneNode** nodeList = obs->getInsideSceneNodeList();
-                for (int o = 0; o < nodeCount; o++)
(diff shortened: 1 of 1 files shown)


macsforme:
If your goal is to make it consistent, why don't you make it consistent the other way, by making the effect visible inside BZW 2.0 objects as well? This would be much less trivial than just ripping out code like you're suggesting.

Zehra:
The 1.0 graphical effect seems to appear quite dated in my opinion.
Perhaps an option to toggle it would be a better solution?

The-Noah:
A toggle would be fine, but personally I like the effect and agree with macsforme.

allejo:
There are quite enough options already in the client; adding yet another toggle would be against the direction of consistency we're trying to go towards, imo.

+1 for making the effect on 2.0 objects.

macsforme:
I never heard one complaint about this effect until you brought it up recently. I don't believe it looks any more dated than the rest of the game. We are also not going to add a toggle for every single obscure effect.

The proposed change is trivial and pointless. If you don't like the effect, you can disable it locally. Otherwise, if you want to work on implementing the effect consistently in other object types, or want to work on a new and improved effect, those would be projects which would receive more consideration. Feel free to reach out to any of the developers for guidance.

JeffM2501:
The IDL is an important part of the game. When inside an object with OO you can't drive backwards. The IDL is there to tell the player they are in that state.

PR #209, opened 2019-07-31
Removal of unused time bomb logic

Zehra:


jwmelto:
Technically, is _wasn't_ unused, but if we want to remove it, that's fine.

Zehra:
Requested changes have been completed:
- Fixed BZFS compile.
- Makefile fixed. (As far as I can tell.)
- Entry in changeLog added.

macsforme:
You didn't remove include/TimeBomb.h from the makefile system. Also, our convention is to use each contributor's real name in the ChangeLog. Since contributing code to the project involves transferring the copyright of the code to the maintainer, there are certain legal formalities that come into play, so it helps to have a record of contributions and the names of contributors.

blast007:
There is also timebomb logic in the configure.ac file that needs removed.

Zehra:
If name presents any issues, I've read both the [Developer Agreement](https://github.com/BZFlag-Dev/bzflag/wiki/Developer-Agreement) and [DEVINFO](https://github.com/BZFlag-Dev/bzflag/blob/2.4/DEVINFO) and agree to all terms and conditions within them.
I would prefer to remain nameless rather than to use my legal name.
If it is not an option, please close this pull request.

allejo:
I'm going to put this on hold for a bit while we discuss our policy about names for attribution since it'd affect some already merged commits.

macsforme:
It was pointed out to me that this is technically just a removal of code and not a contribution of anything new, so the copyright issue is technically moot for this particular case. This resolves my concerns about that issue.

That being said, if you submit anything in the future that actually contains code that you authored, please be prepared to put your name on it. While we don't have a perfect track record of getting the names of all contributors in the past, because of 1) your ever-changing self-portrayal (so that nobody here seems to know who much about you for certain) and 2) at least one example of you going back and trying to scrub old content you contributed, I think we're more than justified in insisting on it in your case to assuage our concerns about future potential copyright-related disputes.

Zehra:
The expectation of copyright formalities for code contributions is reasonable.
The surprise of it being possibly required for code cleanup came as a bit of a shock, so thanks for the clarification.

allejo:
Bumping for @BZFlag-Dev/developers. Is this good to merge?

PR #212, opened 2019-08-11
updated thief control

Zehra:
Minor update which replaces memory managed code with code which performs the same logic and is cleaner.
Two lines do go over 120 characters, but still appear to fall within code guidelines.

### DIFF
--- a/plugins/thiefControl/thiefControl.cpp
+++ b/plugins/thiefControl/thiefControl.cpp
@@ -24,6 +24,8 @@ public:
     }
     virtual void Init( const char* config );
     virtual void Event( bz_EventData *eventData );
+private:
+    void dropThief(bz_FlagTransferredEventData_V1 *data);
 };
 
 BZ_PLUGIN(ThiefControl)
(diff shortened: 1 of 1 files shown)


jwmelto:
What does this change **do**? It appears to be entirely cosmetic. I'm a fan of cosmetic improvements, but why is it worth applying?

Zehra:
It's a minor refactor, so the memory is no longer being manually allocated and deleted.
With the changes, it's allocated and deleted automatically with the _bz_getPlayerTeam_.
In my opinion, it's a simple plug-in and there isn't much reason to do otherwise.
Additionally, some minor changes were done to thiefControl previously within the _[thiefControl in 2.2 leaks memory](https://forums.bzflag.org/viewtopic.php?f=78&t=11388)_ and _[Cleaned up ThiefControl code](https://forums.bzflag.org/viewtopic.php?f=78&t=14520)_ forum threads.

plugins/thiefControl/thiefControl.cpp
```diff
@@ -36,33 +36,20 @@ void ThiefControl::Init(const char* /*config*/ )
...
-    bz_BasePlayerRecord *playerTo = bz_getPlayerByIndex(data->toPlayerID);
-
-    if (!playerTo)
-    {
-        bz_freePlayerRecord(playerFrom);
-        return;
-    }
+	
+    bz_eTeamType thiefTeam = bz_getPlayerTeam(data->toPlayerID);
+    bz_eTeamType targetTeam = bz_getPlayerTeam(data->fromPlayerID);
```
jwmelto: This is the fundamental change, using `bz_getPlayerTeam` instead of `bz_getPlayerByIndex`. While a maintenance burden, the previous code did some explicit error checking for the case where a playerID no long maps to a player record. It may have been pedantic completeness, or it may have been robust. What isn't clear is which, and if this change is equally robust.

Specifically, what does `bz_getPlayerTeam` return in the case where the playerID doesn't map to an existing record?

Beyond that, removing the explicit memory management is a Good Thing™

Zehra: According to "bzfsAPI.cxx", it returns **eNoTeam**, if the _playerID_ does not map to an existing record.

```
BZF_API bz_eTeamType bz_getPlayerTeam( int playerID )
{
    GameKeeper::Player *player = GameKeeper::Player::getPlayerByIndex(playerID);

    if (!player)
        return eNoTeam;

    return convertTeam(player->player.getTeam());
}

jwmelto: So for equivalent functionality, you should test both values for `eNoTeam` and early return if either, right?

Zehra: For precisely equivalent functionality, yes.

macsforme:
To me, this is a pointless change. It appears to do almost exactly the same thing, with just a slightly different method.

I would always encourage contributors to submit pull requests for changes (even small ones) that involve some concrete improvement or bug fix, but I don't see the point in considering a PR for minor refactoring (unless it significantly improves readability or functionality) or for removing features perceived as stale. Time and energy spent reviewing PRs is better spent on more ambitious changes that provide some measurable benefit to the game. The issue tracker lists many examples of good potential projects for a contributor.

I suggest closing this PR.

PR #214, opened 2019-08-20
Add bz_eReloadEvent API event

allejo:
This PR will allow plug-ins to safely and easily hook into `/reload` and `bz_reload*()` events. This will also allow for custom `/reload` subcommands without needing to overload the `/reload` command.

The reason behind this change is to allow plugins to have custom reload functionality.

### DIFF
--- a/include/bzfsAPI.h
+++ b/include/bzfsAPI.h
@@ -338,6 +338,7 @@ typedef enum
     bz_ePermissionModificationEvent,
     bz_eAllowServerShotFiredEvent,
     bz_ePlayerDeathFinalizedEvent,
+    bz_eReloadEvent,
     bz_eLastEvent    //this is never used as an event, just show it's the last one
 } bz_eEventType;
 
@@ -1460,6 +1461,20 @@ public:
     bool customPerm;
 };
 
+class BZF_API bz_ReloadData_V1 : public bz_EventData
+{
+public:
+    bz_ReloadData_V1() : bz_EventData(bz_eReloadEvent)
+        , playerID(-1)
+        , target("all")
+        , handled(false)
+    {}
+
+    int playerID;
+    const char* target;
+    bool handled;
+};
+
 // logging
 BZF_API void bz_debugMessage ( int debugLevel, const char* message );
 BZF_API void bz_debugMessagef( int debugLevel, const char* fmt, ... );
@@ -2163,6 +2178,7 @@ BZF_API void bz_gameOver(int playerID, bz_eTeamType team = eNoTeam);
 BZF_API bool bz_restart ( void );
 BZF_API const char* bz_getServerOwner();
 
+BZF_API void bz_reloadAll();
 BZF_API void bz_reloadLocalBans();
 BZF_API void bz_reloadMasterBans();
 BZF_API void bz_reloadGroups();
--- a/src/bzfs/bzfsAPI.cxx
+++ b/src/bzfs/bzfsAPI.cxx
@@ -4286,10 +4286,29 @@ BZF_API const char* bz_getServerOwner()
     return getPublicOwner().c_str();
 }
 
+BZF_API void bz_reloadAll()
+{
+    bz_ReloadData_V1 event;
+    event.target = "all";
+    worldEventManager.callEvents(bz_eReloadEvent, &event);
+
+    bz_reloadLocalBans();
+    bz_reloadMasterBans();
+    bz_reloadGroups();
+    bz_reloadUsers();
+    bz_reloadHelp();
+    bz_reloadBadwords();
+}
+
 BZF_API void bz_reloadLocalBans()
 {
     // reload the banlist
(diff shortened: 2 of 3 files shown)


src/bzfs/commands.cxx
```diff
@@ -2805,8 +2812,14 @@ bool ReloadCommand::operator() (const char   *message,
     }
     else
     {
+        // Allow a plug-in to take over
```
jwmelto: Assuming the unconditional event manager call is correct, then shouldn't this be moved immediately after it?

I apologize for not being clear on how the event manager interacts; is the point to "do what we do" and allow plugins to extend, or do we allow plugins to override?

src/bzfs/bzfsAPI.cxx
```diff
@@ -4286,10 +4286,29 @@ BZF_API const char* bz_getServerOwner()
...
+}
+
 BZF_API void bz_reloadLocalBans()
 {
     // reload the banlist
     logDebugMessage(3,"Reloading bans\n");
+
+    bz_ReloadData_V1 event;
+    event.target = "bans";
+    worldEventManager.callEvents(bz_eReloadEvent, &event);
```
jwmelto: as noted below, what's the proper interaction with the event manager? Should the call be before the `acl.load()` like this, meaning this might potentially undo a plugin action, or should the event manager call be after the defined action? Or should the event be tested to see if it was handled?

src/bzfs/bzfsAPI.cxx
```diff
@@ -4302,6 +4321,11 @@ BZF_API void bz_reloadMasterBans()
 
     // reload the banlist
     logDebugMessage(3,"Reloading master bans\n");
+
+    bz_ReloadData_V1 event;
+    event.target = "masterbans";
+    worldEventManager.callEvents(bz_eReloadEvent, &event);
```
jwmelto: this method can short-circuit (line 4320 above), meaning the event manager will not dispatch this message. Is that right?

src/bzfs/bzfsAPI.cxx
```diff
@@ -4286,10 +4286,29 @@ BZF_API const char* bz_getServerOwner()
     return getPublicOwner().c_str();
 }
 
+BZF_API void bz_reloadAll()
+{
+    bz_ReloadData_V1 event;
+    event.target = "all";
+    worldEventManager.callEvents(bz_eReloadEvent, &event);
```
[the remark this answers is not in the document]
jwmelto: technically no, since this will send "all" and each call below will send a different string, but if "all" is exploded on the other side then ...

include/bzfsAPI.h
```diff
@@ -1460,6 +1461,20 @@ class BZF_API bz_PermissionModificationData_V1 : public bz_EventData
...
 };
 
+class BZF_API bz_ReloadData_V1 : public bz_EventData
+{
+public:
+    bz_ReloadData_V1() : bz_EventData(bz_eReloadEvent)
+        , playerID(-1)
+        , target("all")
+        , handled(false)
+    {}
```
jwmelto: why is this in-line, and why is the default target `"all"`? Seems like `""` might be a more reasonable default?

allejo: - It's in-line because that's what some of the other events used. Though it appears our data classes are pretty inconsistent. I'm not opposed to switching the format.
- As for the `""`, I don't feel strongly either way. My only reasoning was that running a `/reload` by itself will trigger a reload of everything which is the equivalent of `/reload all`.

src/bzfs/bzfsAPI.cxx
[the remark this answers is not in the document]
allejo: Yea, I was thinking that if we want to listen to an explicit `bz_reloadAll()` or `/reload all`, it'd be useful to know when that happens even though it would trigger subsequent `bz_eReloadEvent`.

src/bzfs/commands.cxx
```diff
@@ -2762,13 +2762,20 @@ bool ReloadCommand::operator() (const char   *message,
 
     logDebugMessage(3,"Command is %s\n", cmd.c_str());
 
+    bz_ReloadData_V1 event;
+    event.playerID = t;
+    event.target = cmd.c_str();
```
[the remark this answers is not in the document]
allejo: good call, this would allow for `event.target` to be `"all"` at the moment, I can fix that

allejo:
@jwmelto since the remaining comments are related, my thinking was that `bz_eReloadEvent` is sent as a notification event, so it'll only be sent when something actually happens (e.g. the masterban bailing out early).

As for the `event.handled`, I was thinking that we shouldn't let plug-ins take over built-in bzfs functionality but, we could allow plug-ins to hook into unknown reload events at which point, bzfs silences its failure messages. I know this goes against my original thought of having this event be notification-only. And having a `handled` could be misleading since you can't take over built-in functionality.

Whether the event should be fired before or after the action, I'm on the fence about.

Thoughts?

PR #215, opened 2019-08-21
Fix missing include

h3xx:
Causes issues in newer versions of GCC.

Broken since aa19b4eb0b064bc8e83b47337916d5caedd0bcac

### DIFF
--- a/src/common/TimeKeeper.cxx
+++ b/src/common/TimeKeeper.cxx
@@ -14,6 +14,7 @@
 #include "TimeKeeper.h"
 
 /* system implementation headers */
+#include <cstring>
 #include <thread>
 
 /* common implementation headers */


macsforme:
Which functions/types require this additional header? The original author and I are both on macOS with Xcode/clang, and I’m not getting any error at the moment.

blast007:
With gcc 6.3.0:
```
TimeKeeper.cxx: In static member function ‘static const char* TimeKeeper::timestamp()’:
TimeKeeper.cxx:99:91: error: ‘strncpy’ was not declared in this scope
                                       now->tm_hour, now->tm_min, now->tm_sec).c_str(), 256);
                                                                                           ^
```

blast007:
The PR has been merged.  Thanks!

PR #222, opened 2019-10-29
Require openGL context 2.1

atupone:
Requesting openGL2.1

Removing all the check for things that are mandatory for that openGL version

### DIFF
--- a/src/bzflag/BackgroundRenderer.cxx
+++ b/src/bzflag/BackgroundRenderer.cxx
@@ -795,12 +795,6 @@ void BackgroundRenderer::setupSkybox()
     for (i = 0; i < 6; i++)
         bzmats[i]->setReference();
 
-    // setup the wrap mode
-    if (GLEW_EXT_texture_edge_clamp)
-        skyboxWrapMode = GL_CLAMP_TO_EDGE;
-    else
-        skyboxWrapMode = GL_CLAMP;
-
     // setup the corner colors
     const int cornerFaces[8][3] =
     {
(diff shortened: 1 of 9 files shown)


src/bzflag/BackgroundRenderer.cxx
```diff
@@ -853,8 +847,8 @@ void BackgroundRenderer::drawSkybox()
     if (!BZDBCache::drawGround)
     {
         tm.bind(skyboxTexID[5]); // bottom
-        glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, skyboxWrapMode);
-        glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, skyboxWrapMode);
+        glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_CLAMP_TO_EDGE);
+        glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE);
         glBegin(GL_TRIANGLE_STRIP);
```
jwmelto: This is begging for a function (or at least a lambda) to reduce the amount of copy/paste

src/bzflag/bzflag.cxx
```diff
@@ -1288,7 +1288,7 @@ int         main(int argc, char** argv)
             case GL_OUT_OF_MEMORY:
                 std::cerr << "ERROR: GL_OUT_OF_MEMORY" << std::endl;
                 break;
-#ifdef GL_VERSION_1_2
+#ifdef GL_TABLE_TOO_LARGE
```
jwmelto: is this optional in GL 2.1? If not, we can probably do away with the `#ifdef`

atupone: That is eventually for GLES. GL_TABLE_TOO_LARGE is not applicable to it

atupone:
Before merging I would like to see if some of the code that I removed still apply to the eventual GLES porting.

src/bzflag/bzflag.cxx
macsforme: When that time comes, we should find a better test than that (and there will need to be many others). Compiler errors will show any unsupported code that needs to be updated. In my GLES port, I just defined `HAVE_GLES`, which also made it easy to find any GLES-specific code. I suggest just remove that `#ifdef` for now.

atupone: lot of code out of here is using that ifdef.
But I don't care removing that for now

src/bzflag/BackgroundRenderer.cxx
atupone: It seems the pattern for introducing a 3d texture there

PR #223, opened 2019-10-31
unconditionally enable Blending

atupone:
Remove blend from the option menu.
Consider blend always true in the code.

### DIFF
--- a/include/SceneNode.h
+++ b/include/SceneNode.h
@@ -51,7 +51,6 @@
 #define myColor4f(r, g, b, a)   SceneNode::glColor4f(r, g, b, a)
 #define myColor3fv(rgb)     SceneNode::glColor3fv(rgb)
 #define myColor4fv(rgba)    SceneNode::glColor4fv(rgba)
-#define myStipple(alpha)    SceneNode::setStipple(alpha)
 
 class ViewFrustum;
 class SceneRenderer;
--- a/src/bzflag/DisplayMenu.cxx
+++ b/src/bzflag/DisplayMenu.cxx
@@ -55,16 +55,6 @@ DisplayMenu::DisplayMenu() : formatMenu(NULL)
 
     option = new HUDuiList;
     option->setFontFace(fontFace);
-    option->setLabel("Blending:");
-    option->setCallback(callback, "2");
-    options = &option->getList();
-    options->push_back(std::string("Off"));
-    options->push_back(std::string("On"));
-    option->update();
-    listHUD.push_back(option);
-
-    option = new HUDuiList;
-    option->setFontFace(fontFace);
     option->setLabel("Smoothing:");
     option->setCallback(callback, "3");
     options = &option->getList();
@@ -324,7 +314,6 @@ void            DisplayMenu::resize(int _width, int _height)
     {
         TextureManager& tm = TextureManager::instance();
         ((HUDuiList*)listHUD[i++])->setIndex(BZDB.isTrue("dither"));
-        ((HUDuiList*)listHUD[i++])->setIndex(BZDB.isTrue("blend"));
         ((HUDuiList*)listHUD[i++])->setIndex(BZDB.isTrue("smooth"));
         if (BZDBCache::lighting)
         {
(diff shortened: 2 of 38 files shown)


include/SceneNode.h
```diff
@@ -51,7 +51,6 @@
 #define myColor4f(r, g, b, a)   SceneNode::glColor4f(r, g, b, a)
 #define myColor3fv(rgb)     SceneNode::glColor3fv(rgb)
 #define myColor4fv(rgba)    SceneNode::glColor4fv(rgba)
-#define myStipple(alpha)    SceneNode::setStipple(alpha)
```
jwmelto: stipple wasn't removed above; are you sure you meant to do this?

src/bzflag/DisplayMenu.cxx
```diff
@@ -404,10 +393,6 @@ void            DisplayMenu::callback(HUDuiControl* w, const void* data)
         BZDB.set("dither", list->getIndex() ? "1" : "0");
         sceneRenderer->notifyStyleChange();
         break;
-    case '2':
```
jwmelto: you shouldn't just remove case 2. Sadly, there is no default case in this switch. Either add a default case (which is best) or an empty or error case for 2.

src/bzflag/playing.cxx
```diff
@@ -3625,7 +3625,7 @@ bool            addExplosion(const float* _pos,
     if (sceneRenderer->useQuality() < 1) return false;
 
     // don't add explosion if blending or texture mapping are off
```
jwmelto: change the comment to reflect the code.

src/geometry/BoltSceneNode.cxx
```diff
@@ -159,33 +159,23 @@ void            BoltSceneNode::addLight(
...
-        builder.setAlphaFunc();
-        if ((RENDERER.useQuality() >= 3) && drawFlares)
-        {
-            builder.setShading(GL_SMOOTH);
-            builder.enableMaterial(false);
-        }
-        else
-            builder.setShading(texturing ? GL_FLAT : GL_SMOOTH);
-    }
+    const int shotLength = (int)(BZDBCache::shotLength * 3.0f);
```
jwmelto: C-style casts are evil. Either use a C++ cast or a "function-style" cast (anonymous temporary:
```c++
const int shotLength = static_cast<int>(BZDBCache::shotLength * 3.0);
const int shotLength = int( BZDBCache::shotLength * 3.0 );
```
(although I don't know why we are multiplying by a float anyway)
(and magic numbers like "3.0" are maintenance burdens)
(and I realize you just removed the `if` but any time you touch a line of code, it's nice to leave it better)

include/SceneNode.h
atupone: setStipple is used, myStipple is no more

src/bzflag/DisplayMenu.cxx
atupone: I agree! Doing it

src/bzflag/playing.cxx
atupone: OK

src/geometry/BoltSceneNode.cxx
atupone: I don't know why we multiply it either.
Removing

jwmelto: I wasn’t suggesting the multiplication was bad, just that multiplying by `3.0f` vs `3` was silly, especially given the conversion to `int`

atupone: I re-added it

macsforme: > I wasn’t suggesting the multiplication was bad, just that multiplying by `3.0f` vs `3` was silly, especially given the conversion to `int`

I just wanted to point out that the result is not necessarily the same (if the value has a fractional part).

jwmelto: Not to quibble but it is the same. `1.1 * 3.0 == 1.1 * 3`. Integer and floating point numeric operations are well defined and well behaved

macsforme: Okay I see what you were saying. I thought you were suggesting to do the multiplication after the cast.

PR #224, opened 2019-11-13
Convert code to use glm

atupone:
This is using glm vec2/vec3/vec4 instead of GLfloat[2]/[3]/[4] where it is needed for the openGL2.0 porting.
More can/could be done. 
It could be also used for physics but is not done, unless some want to spend some time on it


macsforme:
Is this basically a continuation of #194? If so then I think it makes sense to expand on that, although I'm curious why you say this is needed for OpenGL 2.0 porting?

atupone:
That PR you refer to what, to me, is the least we can do to get rids of vectors and vectors_old.
This is instead the start of my porting.
Sure it is not needed for openGL 2.0, but the road I chosen for my porting, is heavily based on glm.
From my commits I separated the translation from vector to glm that was in, and put these in only one commit. This!
I did not substitute every triplet or quadruple 'cause I had problem with meshes implementation. Maybe if someone can invest time on it the code could be cleaner.

src/bzflag/BackgroundRenderer.cxx
```diff
@@ -851,18 +852,18 @@ void BackgroundRenderer::drawSkybox()
         glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE);
         glBegin(GL_TRIANGLE_STRIP);
         {
-            glTexCoord2fv(txcds[0]);
+            glTexCoord2fv(glm::value_ptr(txcds[0]));
```
macsforme: Any particular reason to use glm vectors for the vertex and texture coordinates, but not for the color component, whereas you do use a glm vector for color components in other files?

atupone: No reason. Only they did not help me later. I think we should always use glm vectors but need some help 'cause is a lot of work and test

include/MeshPolySceneNode.h
```diff
@@ -54,7 +56,7 @@ class MeshPolySceneNode : public WallSceneNode
                  const GLfloat3Array& vertices,
                  const GLfloat3Array& normals,
                  const GLfloat2Array& texcoords,
-                 const GLfloat* normal);
+                 const glm::vec3 normal);
```
jwmelto: note the (unintentional?) pass by value here

include/SceneRenderer.h
```diff
@@ -107,7 +108,7 @@ class SceneRenderer : public Singleton<SceneRenderer>
 
     void      clearRadar(float opacity);
 
-    void      getGroundUV(const float p[2], float uv[2]) const;
+    void      getGroundUV(const glm::vec2 p, glm::vec2 uv) const;
```
jwmelto: will this work? `uv` appears to be pass-by-value, but I suspect the method is trying to populate it.

src/geometry/TankSceneNode.cxx
```diff
@@ -19,8 +19,14 @@
 #include "TankSceneNode.h"
 
 // system headers
+#define GLM_ENABLE_EXPERIMENTAL
```
jwmelto: This should be before **any** includes. As a rule, I'm not a fan of defining macros that affect subsequent include files in the code, because you don't know what has been included that may nullify your intended effect. Rather, these types of defines should be specified on the compilation command to ensure they have the intended effect.

include/EighthDBaseSceneNode.h
```diff
@@ -34,7 +37,7 @@ class EighthDBaseSceneNode : public EighthDimSceneNode
     {
     public:
         EighthDBaseRenderNode(const EighthDBaseSceneNode *,
-                              const float pos[3],
+                              const glm::vec3 pos,
                               const float size[3], float rotation);
```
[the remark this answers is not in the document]
macsforme: +1 on passing by reference to avoid the copy... good catch.

atupone:
Well. I  close this PR and do the same things (partially when needed) on the other commits. I would try to apply the comments there

PR #227, opened 2019-11-16
Enable dithering always

atupone:


macsforme:
Doesn't this technically enable dithering always, since the initial value of GL_DITHER is GL_TRUE? Still the desired result, I think, but I just wanted to clarify what this is doing.

atupone:
src/ogl/OpenGLGState.cxx changes it to disable

atupone:
But I understand that dithering has no effect if the color buffer has enough bits.

macsforme:
> tupone: as for the dithering in bzflag, it was disabled when drawing the sky. I don't know if it was for performance or for other unwanted effects. Someone should test it

I did a quick test making the sky respect the in-game dithering setting. I could not reliably measure a difference in framerate on my system. If there is a difference, it seemed to be in the low tenths of a percent reduction.

I think this can be merged.

atupone:
Dithering is a (mostly) noop when using a 24 bits color buffer.
Did you test with lower resolution (like 8 bits color buffer)?

macsforme:
macOS has no way to emulate 265 colors that I can find. If we want to test that then someone else will have to do it.

PR #229, opened 2019-11-29
Add bz_getSpawnPointWithin to API

allejo:
When I originally introduced `bz_getRandomPoint()` to the API, I didn't take into consideration that there may be obstacles in said zones. This is what `bz_getSpawnPointWithin()` is supposed to solve by using our `DropGeometry::dropPlayer()`.

~However, I don't have a fail-safe right now so this can potentially introduce an infinite loop if there are no safe locations. Should I use `_spawnMaxCompTime` to decide when to exit out like our spawn policies? What behavior should exist if this exits out?~

Or is there a better solution to this problem entirely?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,7 @@
 BZFlag 2.4.19
 -------------
 
+* Add bz_getSpawnPointWithin to API - Vladimir Jimenez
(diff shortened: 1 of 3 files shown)


jwmelto:
This is outside of my wheel house

PR #236, opened 2019-12-27
C++17

jwmelto:
This PR updates our minimum language compliance to C++14, with optional support for C++17.


README.macOS
```diff
@@ -20,3 +20,55 @@ Building BZFlag from Source
...
+ld: warning: text-based stub file /System/Library/Frameworks/CoreServices.framework/Versions/A/Frameworks/LaunchServices.framework/Versions/A/LaunchServices.tbd and library file /System/Library/Frameworks/CoreServices.framework/Versions/A/Frameworks/LaunchServices.framework/Versions/A/LaunchServices are out of sync. Falling back to library file for linking.
+ld: warning: text-based stub file /System/Library/Frameworks/CoreServices.framework/Versions/A/Frameworks/SharedFileList.framework/Versions/A/SharedFileList.tbd and library file /System/Library/Frameworks/CoreServices.framework/Versions/A/Frameworks/SharedFileList.framework/Versions/A/SharedFileList are out of sync. Falling back to library file for linking.
+
+HUDRenderer.cxx:1117:13: warning: 'gluProject' is deprecated: first deprecated in macOS 10.9 - "Use GLKMathProject" [-Wdeprecated-declarations]
+HUDRenderer.cxx:1260:5: warning: 'gluProject' is deprecated: first deprecated in macOS 10.9 - "Use GLKMathProject" [-Wdeprecated-declarations]
+
+ShockWaveStrategy.cxx:27:50: warning: field 'radius' is uninitialized when used here [-Wuninitialized]
```
atupone: Hope this is not more necessary

jwmelto: I had those warnings before changing to C++14. They may also be an indication of instability in my environment. If the collective thinks that file should be reverted, I’m okay with it, and ultimately all warnings need to be resolved but that’s outside the scope of this PR.

atupone: to be more clear: 

my last commit "Wrong Initialization order" should have fixed that.

I did not have this warning on linux, So I cannot say for sure, but if is is you can remove this line from the README.macOS

jwmelto: Yes, I pulled from master and confirmed that the warning is now gone. All that is left are the GLU deprecations.

atupone:
As for the new commit about gluProject moved to glm:project, I hope you can remove from README.macos the warning about HUDRenderer.cxx

configure.ac
```diff
@@ -217,18 +217,11 @@ AC_PROG_CXX
...
-    elif test "x$ax_cv_cxx_compile_cxx0x_gxx" = xyes; then
-	CXXFLAGS="$CXXFLAGS -std=gnu++0x"
-	user_CXXFLAGS="$user_CXXFLAGS -std=gnu++0x"	# for --enable-debug
-    else
-	AC_MSG_ERROR([A c++ compiler with C++0x support is required to build BZFlag])
-    fi
-fi
+
+# C++14 is the minimum standard. C++17 isn't widely available
+#AX_CXX_COMPILE_STDCXX([11], [noext])
```
allejo: If we're moving to C++14 being the new minimum, do we need C++11 commented out?

jwmelto: I left it for comparison but I’m happy to remove it.

allejo: I don't feel strongly either way about this, I was more so curious about its purpose

README.macOS
```diff
@@ -20,3 +20,12 @@ Building BZFlag from Source
 ---------------------------
 BZFlag builds on macOS are now done using premake5 and Xcode 7 (or later).
 Please refer to the README.premake5 file for build instructions.
+
+
+Known Issues
+------------
+The following warnings can be expected in a MacOS build (for now):
```
macsforme: From a context of historical tendencies in our code/documentation, I would strongly prefer that we do not document issues in this manner. Generally speaking, such comments tend to be forgotten about, the context is lost, and we end up with TODOs and FIXMEs littered throughout the code. Regarding READMEs specifically, I would prefer that they generally be set-it-and-forget-it, rather than having some kind of working list of issues or deficiencies. I am also working on resolving these last two warnings, after which our code should compile cleanly on macOS.

atupone:
who fixed all the glu warnings? I fixed only the gluProject

configure.ac
jwmelto: It was there in the first place because I wanted to verify that it compiled that way (C++11) vs the old config as a sanity check. As evidenced by the branch name, I intended to go all the way to C++17, but I found the Mac compiler (at least) has insufficient C++17 support). At one time I had flags for all 3 versions. As I tested, I dropped down to just the one. Say the word and this comment is gone.

README.macOS
jwmelto: Again, I don’t mind reverting this file entirely. My purpose in adding this section was to inform a Mac developer of “expected” warnings so they would know they hadn’t done anything wrong.

macsforme:
> who fixed all the glu warnings? I fixed only the gluProject

They were suppressed in 82ced1b2. It requires running premake again for this to take effect, though.

jwmelto:
> As far as whether we actually want to change our C++ standard, that's a decision I'd like to see more discussion about, along with some data on operating system and compiler support. Keep in mind that we haven't even had an official release that fully utilizes C++11, yet.

I figured this PR would be a forcing function for any discussion, if needed. What form would you like this discussion to take? As to support, clang 10 (MacOS), g++ 4.9, and VS 2017 all support C++14.

blast007:
It looks like [some parts of C++14][1] need gcc 5 (or maybe even a [newer][2] version?).  Even in 4.9, some features of C++11 were added, such as [support for <regex>][3] (see under "Runtime Library (libstdc++)").  I imagine the situation is the same with other compilers.  So maybe the conversation should be around which features of newer C++ standards we want to use (not just a blanket "C++14" or "C++17") and which platforms we need to compile on (such as Debian Buster, Visual Studio 2017, macOS 10.13 w/ Xcode 10.1, etc).

[1]: https://gcc.gnu.org/projects/cxx-status.html#cxx14
[2]: https://gcc.gnu.org/onlinedocs/gcc-5.5.0/libstdc++/manual/manual/status.html#status.iso.2014
[3]: https://gcc.gnu.org/gcc-4.9/changes.html#cxx

jwmelto:
> So maybe the conversation should be around which features of newer C++ standards we want to use (not just a blanket "C++14" or "C++17") and which platforms we need to compile on (such as Debian Buster, Visual Studio 2017, macOS 10.13 w/ Xcode 10.1, etc).

Exactly. Sort of. Rather than getting into a pick-and-choose feature set (although, for what it's worth CMake makes specifying these features easy), I think it is better to choose a minimum compiler version. I would note that our DEVINFO does say that extensions should not be used, so we should require standards conformance and allow developers to write to the standard.

We specify versions of things required to build this game (SDL2, for example). We require people to install things (like an alpha version of premake5). We **can** specify minimum supported compiler. gcc is up to version 9.2; the last 4.9 release was over 3 years ago.

Debian Buster has a [gcc-7](https://packages.debian.org/buster/gcc-7) (7.4) package. Windows [VS 2017](https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance) supports the C++14 flag (VS 2019 is out). MacOS 10.13 [XCode 10.1](https://developer.apple.com/documentation/xcode_release_notes/xcode_10_release_notes) [clang 10](https://clang.llvm.org/cxx_status.html) claims full C++14 compliance. Actually, [this page](https://en.cppreference.com/w/cpp/compiler_support) is probably the most useful, with respect to compliance.

I guess the question is, for a development head (**master**) do we really need to lock into the lowest common denominator at this point? I'm not sure I see any reason to avoid adopting a 5-year old standard, but maybe I'm missing something?

Issue #201, opened 2019-07-03
bzflag is not working properly with 3+ monitors


jose1711:
with 3 monitor setup `bzflag` window flashes and mouse cursor jumps making it really hard to even exit the game. this video illustrates the issue: https://www.youtube.com/watch?v=HTjVPt2RnXs. once all but one displays are disabled (using `xrandr`) `bzflag` can be played.

blast007:
What Linux distribution and distribution version are you running?  What video card and driver/version?  Do all three of your screens have the same resolution?  Does it also happen with two displays or just with three?

jose1711:
> What Linux distribution and distribution version are you running? 

Arch Linux current

> What video card and driver/version?

Nvidia GeForce GTX 750Ti, proprietary driver 430.26

> Do all three of your screens have the same resolution?

no.

```
$ xrandr | grep connected
VGA-0 connected 1920x1080+0+0 (normal left inverted right x axis y axis) 509mm x 286mm
DVI-D-0 connected primary 1280x1024+1920+0 (normal left inverted right x axis y axis) 376mm x 301mm
HDMI-0 connected 1440x900+3200+0 (normal left inverted right x axis y axis) 428mm x 255mm

```

> Does it also happen with two displays or just with three?

I disabled `VGA-0` output reducing number of displays to 2 and the issue disappeared.

macsforme:
I have a three-monitor setup as well, comprised of (from left to right) a 1080p monitor in portrait mode, and then two 1080p monitors in landscape mode. In Ubuntu Linux 19.04, when I use the BZFlag version from the snap repository, which uses SDL 1.2 (probably the same as what is in the repositories for other distributions), the window is spread across all three monitors and is very zoomed in with a large portion of the window missing. When I toggle to windowed mode, display options shows this:

![Screenshot from 2019-07-10 23-48-00](https://user-images.githubusercontent.com/4955448/61029773-a0be4700-a370-11e9-9001-50b2d85e63aa.png)

When I compile from source against SDL 2, the window occupies my main monitor only, with the proper resolution and layout, and display options shows this:

![Screenshot from 2019-07-10 23-57-59](https://user-images.githubusercontent.com/4955448/61029821-c9464100-a370-11e9-999f-4a84de952040.png)

This leads me to wonder if this is an SDL 1.2 issue.

@jose1711, can you try compiling the game from source using SDL 2, and tell us if you have a different result? `./configure -help` shows the options needed to do that.

[...]

PR #238, opened 2019-12-29
SDL2: Only create the window once

blast007:
This changes SDL2Window so that it only creates a window once.  It initially creates a window at the base size (640x480 or what was provided by the -window argument).  If the game is to be run fullscreen, it then changes to fullscreen.  Window size or resolution changes and switching between fullscreen and windowed merely changes the size and properties of the existing window instead of destroying and recreating the window.

This is similar to #228 but takes it further by never recreating the window.  Additionally, by creating the initial window at a lower resolution of 640x480, it works around the SDL2 multi-monitor issue in #201, assuming that there's at least a usable area of 640x480 (plus the size of the window decorations) on the primary display.

This likely breaks non-SDL2 platform code as I did adjust a couple other files.  I have tested this mostly on Linux.  I have not tested on macOS or with SDL1 or native Linux platform code.

I ran a brief test on Windows and changing the fullscreen resolution is broken, so I'll need to look into that some more.  Specifically, when I had two 2560x1440 displays, was running fullscreen, and tried to change to 1920x1080, it started rendering on the secondary display with an old snapshot of the menu stuck on the primary display.

### DIFF
--- a/src/bzflag/MainWindow.cxx
+++ b/src/bzflag/MainWindow.cxx
@@ -277,10 +277,9 @@ void MainWindow::setProjectionRadar(int x, int y, int w, int h, float radarRange
 
 void            MainWindow::resize()
 {
-    window->getSize(trueWidth, trueHeight);
-    window->makeCurrent();
     if (!window->create())
         faulting = true;
+    window->getSize(trueWidth, trueHeight);
     setQuadrant(quadrant);
 }
 
--- a/src/platform/SDL2Window.cxx
+++ b/src/platform/SDL2Window.cxx
@@ -125,13 +125,7 @@ void SDLWindow::setSize(int _width, int _height)
 
 void SDLWindow::getSize(int& width, int& height) const
 {
-    if (fullScreen)
-        const_cast<SDLDisplay *>((const SDLDisplay *)getDisplay())->getWindowSize(width, height);
-    else
-    {
-        width  = base_width;
-        height = base_height;
-    }
+    SDL_GL_GetDrawableSize(windowId, &width, &height);
 }
 
 void SDLWindow::setGamma(float gamma)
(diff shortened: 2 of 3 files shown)


macsforme:
On macOS, these work:
- Toggling fullscreen via F1
- Maximizing and un-maximizing via the window title bar
- Resizing the window via the bottom right corner

These do not work:
- Adjusting the fullscreen resolution via the menu, as you mentioned
  - The modeset appears to work, but then the content is partially cut off on the screen, and it appears that you can't set the original resolution again once you change it
- Toggling to fullscreen via F1 after maximizing via the toolbar
  - The screen just goes black until you press F1 again
  - This sequence of events doesn't really do anything; the game just needs to not choke when it happens

src/bzflag/MainWindow.cxx
```diff
@@ -277,10 +277,9 @@ void MainWindow::setProjectionRadar(int x, int y, int w, int h, float radarRange
 
 void            MainWindow::resize()
 {
-    window->getSize(trueWidth, trueHeight);
-    window->makeCurrent();
     if (!window->create())
         faulting = true;
+    window->getSize(trueWidth, trueHeight);
```
jwmelto: it's not at all clear what this change does. Some comments would be awesome

src/platform/SDL2Window.cxx
```diff
@@ -187,128 +181,110 @@ void SDLWindow::swapBuffers()
...
-    getSize(targetWidth, targetHeight);
-    SDL_bool windowWasGrabbed = SDL_FALSE;
-    if (windowId != NULL)
-        windowWasGrabbed = SDL_GetWindowGrab(windowId);
-
-    // if we have an existing identical window, go no further
-    if (windowId != NULL)
+    // If fullscreen, the target size will be that of the selected fullscreen resolution
+    if (fullScreen)
+        const_cast<SDLDisplay *>((const SDLDisplay *)getDisplay())->getWindowSize(targetWidth, targetHeight);
```
jwmelto:
```suggestion
        const_cast<SDLDisplay *>(static_cast<const SDLDisplay *>(getDisplay()))->getWindowSize(targetWidth, targetHeight);
```
it's odd to mix C-style (bad) and C++-style (less bad) casts in a single statement.

And `getWindowSize` isn't a const method? Why not?

src/platform/SDL2Window.cxx
```diff
@@ -187,128 +181,110 @@ void SDLWindow::swapBuffers()
 bool SDLWindow::create(void)
 {
     int targetWidth, targetHeight;
```
jwmelto: From a complexity standpoint, uninitialized variables are bad. If this block were refactored so that these variables were initialized to the base dimensions, and overridden if `fullScreen` is true (if case below, lose the else), then you wouldn't have uninitialized variables dangling.

src/platform/SDL2Window.cxx
```diff
@@ -187,128 +181,110 @@ void SDLWindow::swapBuffers()
...
-                targetWidth == currentWidth && targetHeight == currentHeight)
-            return true;
+        targetWidth = base_width;
+        targetHeight = base_height;
     }
 
-    // destroy the pre-existing window if it exists
-    if (windowId != NULL)
+    // Create the SDL window if it doesn't already exist
+    if (!windowId)
```
jwmelto: I don't know if this is just a personal preference or if there is research to back this position, but I prefer the explicit test that was there before (updated for modern C++)
```suggestion
    if (windowId != nullptr)
```

src/platform/SDL2Window.cxx
```diff
@@ -187,128 +181,110 @@ void SDLWindow::swapBuffers()
...
+                       SDL_WINDOWPOS_UNDEFINED,
+                       base_width,
+                       base_height,
+                       flags);
+
+        // If the window could not be created, bail out
+        if (!windowId)
+        {
+            printf("Could not create the window: %s\n", SDL_GetError());
+            return false;
```
jwmelto: what happens in this case? There is no window, can the game be played? Multiple returns are bad. 

I wonder if this might be one of those exceptional conditions where throwing an exception is warranted?

src/platform/SDL2Window.cxx
```diff
@@ -187,128 +181,110 @@ void SDLWindow::swapBuffers()
...
+        }
 
-    windowId = SDL_CreateWindow(
-                   title.c_str(),
-                   SDL_WINDOWPOS_UNDEFINED,
-                   SDL_WINDOWPOS_UNDEFINED,
-                   targetWidth,
-                   targetHeight,
-                   flags);
+        // Store the gamma immediately after creating the first window
```
jwmelto: the first window? I thought the point was to only create one.

src/platform/SDL2Window.cxx
```diff
@@ -187,128 +181,110 @@ void SDLWindow::swapBuffers()
...
+            {
+                printf("Unable to set display mode: %s", SDL_GetError());
+                return false;
+            }
         }
+        // Otherwise just set the window size
         else
-            eventStack.push_back(thisEvent);
+        {
+            //SDL_SetWindowSize(windowId, targetWidth, targetHeight);
```
jwmelto: why are we keeping commented out code?

src/platform/SDL2Window.cxx
macsforme: You want to set these two variables and then immediately set them a second time every time this function is called for a fullscreen window?

jwmelto:
```c++
int targetWidth;
```
declares an uninitialized variable.
```c++
int targetWidth = base_width;
```
declares an initialized variable.

We aren't talking wasting CPU cycles or potential for optimization here; we are talking about removing the complexity of an uninitialized variable. You are going to do the assignment in the full screen case regardless.

src/platform/SDL2Window.cxx
macsforme: // set fullscreen again so MainWindow object knows it's full screen
    if (useFullscreen)
        // this will also call window create
        pmainWindow->setFullscreen();
    else
        window->create();
The return value is ignored.

It looks like he was just following the existing pattern. We do need a way to bail out from errors like this, but since that involves cleanup inside SDL (e.g.,` SDL_Quit()`) and possibly outside of SDL as well, that seems more like a separate project entirely. SDL has functions to display an error dialog, or if SDL initialization itself failed, we can always fall back to stderr output. Yet another thing that would be simplified by going all in on SDL.

src/platform/SDL2Window.cxx
macsforme: The opposite, I think, since the purpose of the comparison is now to create a new window versus destroying an existing one.

jwmelto: Yes, of course. Hence the improved maintainability of the explicit logic test.

macsforme:
This fix was reworked into commit 22c452d6 of PR #251.  Closing.

PR #239, opened 2020-01-02
Redo setView with glm

atupone:
Using glm to recompute Model View Matrix


atupone:
not good now :(
I will reopen when is ok

PR #240, opened 2020-01-03
Rewrite Frustum using glm

atupone:
Rewrote Frustum.* using glm
The Model View Matrix has been rebuilt with a single glm method (glm::lookAt)
I cannot find in glm the bzflag projection matrix as it is. Maybe it is wrong in bzflag?
For now I keep that as it was before

### DIFF
--- a/include/Frustum.h
+++ b/include/Frustum.h
@@ -17,7 +17,14 @@
 #ifndef BZF_FRUSTUM_H
 #define BZF_FRUSTUM_H
 
+// Before everything
 #include "common.h"
+
+// System headers
+#include <glm/vec3.hpp>
+#include <glm/vec4.hpp>
+#include <glm/mat4x4.hpp>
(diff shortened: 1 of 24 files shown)


macsforme:
There is a perspective projection matrix formula based on a vertical FOV and one based on a horizontal FOV. I forget which is which, but when I was trying to replicate it a while back, I found that BZFlag uses the opposite one from the most common formula found in online tutorials, etc.

atupone:
> There is a perspective projection matrix formula based on a vertical FOV and one based on a horizontal FOV. I forget which is which, but when I was trying to replicate it a while back, I found that BZFlag uses the opposite one from the most common formula found in online tutorials, etc.

I cannot find this
projectionMatrix[2][1] = -fracHeight;
online I only see 
projectionMatrix[2][1] = 0;

atupone:
Merged

PR #243, opened 2020-01-23
Menu improvements

blast007:
This improves the organization of the menus in the game.  I tried to group menu options better and include breaks between groups of options.  It also fixes the wireframe debug rendering mode that has been broken for quite some time.

The Xcode project will need to have the AudioMenu.cxx/h files removed.


allejo:
Overall I really like this re-organization. It makes a lot more sense. Here are my notes:

- `Input Settings` > `Joystick device`

  Is there a way to gray/disable joystick options when there are no joysticks plugged in? Seems weird to have options you can change that won't have any effect.

  Or at least, thoughts on saying "None detected" instead of "Off" when there are no joysticks plugged in?

- `Display Settings` > `Change Video Format`

  Thoughts on having this be `Change Resolution`?

- `GUI Settings`

  There's no standardization in the "Control Panel" section. There's "Console", "Panel", "ControlPanel", and "Control panel" all used differently

blast007:
I'll mess around with the joystick stuff a bit.

I had (mostly?) left the names of options and values the same for now because otherwise the language files would need to be updated.  I'd like to do that as a separate PR to better track those changes.

allejo:
I'm cool with tracking the content changes in a separate issue/PR so we can take localization into account.

src/bzflag/DisplayMenu.cxx
```diff
@@ -134,44 +136,86 @@ DisplayMenu::DisplayMenu() : formatMenu(NULL)
...
+    option->setCallback(callback, "3");
+    options = &option->getList();
+    options->push_back(std::string("Off"));
+    options->push_back(std::string("On"));
+    option->update();
+    listHUD.push_back(option);
+
+
+
+    if (((BzfWindow*)getMainWindow()->getWindow())->hasGammaControl())
```
jwmelto: C-style casts are evil. If you must cast, prefer "ugly" C++ casts (better to expose the proper interfaces on high-level objects)
```suggestion
    if ( dynamic_cast<BzfWindow*>(getMainWindow()->getWindow())->hasGammaControl())
```

src/bzflag/DisplayMenu.cxx
```diff
@@ -442,8 +473,8 @@ void            DisplayMenu::callback(HUDuiControl* w, const void* data)
     }
     case 'A':
```
jwmelto: you removed several options but their callback code remains (e.g., '1' : dithering). Only the cases that can be executed should remain.

src/bzflag/DisplayMenu.cxx
```diff
@@ -323,9 +331,14 @@ void            DisplayMenu::resize(int _width, int _height)
     if (renderer)
     {
         TextureManager& tm = TextureManager::instance();
-        ((HUDuiList*)listHUD[i++])->setIndex(BZDB.isTrue("dither"));
-        ((HUDuiList*)listHUD[i++])->setIndex(BZDB.isTrue("blend"));
-        ((HUDuiList*)listHUD[i++])->setIndex(BZDB.isTrue("smooth"));
+
+        // Quality
+        ((HUDuiList*)listHUD[i++])->setIndex(renderer->useQuality());
```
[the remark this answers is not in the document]
jwmelto: OK, I see that this is an incredibly brittle coupling between the order of creation in the ctor and here. At the very least, this coupling should be documented, as in "hey make sure the order of access here matches the order of creation"

Probably outside the scope of what you were trying to accomplish, but I'm wondering why these mutation (`setIndex`) operations are happening here in `resize()`. It would be cleaner if the index values were set when the items are constructed.

src/bzflag/DisplayMenu.cxx
```diff
@@ -357,29 +358,59 @@ void            DisplayMenu::resize(int _width, int _height)
...
+        // Multi-Samping
+        if (OpenGLGState::getMaxSamples() > 1)
+        {
+            // multisampling
+            int multisampleLevel = BZDB.evalInt("multisample");
+            int multisampleIndex = 0;
+            while (multisampleLevel > pow(2, multisampleIndex))
+                multisampleIndex++;
+            ((HUDuiList*)listHUD[i++])->setIndex(multisampleIndex);
+        }
```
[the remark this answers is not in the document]
jwmelto: no, I guess not, since the option doesn't get created unless this condition is true

src/bzflag/GUIOptionsMenu.cxx
```diff
@@ -366,7 +347,7 @@ void            GUIOptionsMenu::resize(int _width, int _height)
 
     // use a big font for title, smaller font for the rest
     const float titleFontSize = (float)_height / 15.0f;
-    const float fontSize = (float)_height / 65.0f;
+    const float fontSize = (float)_height / 70.0f;
```
jwmelto: is there anything to be gained by being consistent? In **effectsMenu** this is 60.

src/bzflag/GUIOptionsMenu.cxx
```diff
@@ -388,65 +369,81 @@ void            GUIOptionsMenu::resize(int _width, int _height)
     {
         listHUD[i]->setFontSize(fontSize);
         listHUD[i]->setPosition(x, y);
-        y -= 1.0f * h;
+        if (i == 3 || i == 11 || i == 22)
```
jwmelto: document the magic numbers?

src/bzflag/GUIOptionsMenu.cxx
```diff
@@ -388,65 +369,81 @@ void            GUIOptionsMenu::resize(int _width, int _height)
...
     }
 
     // load current settings
     SceneRenderer* renderer = getSceneRenderer();
     if (renderer)
     {
         int i = 1;
-        ((HUDuiList*)listHUD[i++])->setIndex(BZDBCache::radarStyle);
+        // Scoreboard Font Size
+        ((HUDuiList*)listHUD[i++])->setIndex(static_cast<int>(BZDB.eval("scorefontsize")));
```
jwmelto: same comments on brittle coupling apply here.

src/bzflag/InputMenu.cxx
```diff
@@ -279,15 +288,26 @@ void            InputMenu::callback(HUDuiControl* w, const void* data)
...
+        const bool clamped = (index == 2);
         BZDB.set("mouseClamp", clamped ? "1" : "0");
     }
     break;
 
+
+    case 'M':
+    {
+        SceneRenderer* renderer = getSceneRenderer();
+        renderer->setMaxMotionFactor(listHUD->getIndex() - 11);
```
jwmelto: elsewhere, `renderer` is tested for `nullptr` before use. Is it guaranteed not to be null here?

src/bzflag/InputMenu.cxx
```diff
@@ -325,15 +345,18 @@ void            InputMenu::resize(int _width, int _height)
...
+    x = 0.5f * ((float)_width);
     y -= 0.6f * titleHeight;
     const float h = fm.getStrHeight(MainMenu::getFontFace(), fontSize, " ");
     const int count = listHUD.size();
     for (i = 1; i < count; i++)
     {
         listHUD[i]->setFontSize(fontSize);
         listHUD[i]->setPosition(x, y);
-        y -= 1.0f * h;
+        if (i == 1 || i == 2 || i == 7 || i == 9)
```
jwmelto: more magic numbers

src/bzflag/InputMenu.cxx
```diff
@@ -346,6 +369,10 @@ void            InputMenu::resize(int _width, int _height)
     }
     if (BZDB.isTrue("allowInputChange"))
         activeInput->setIndex(0);
+
+    i = 9;
+    SceneRenderer* renderer = getSceneRenderer();
+    ((HUDuiList*)listHUD[i++])->setIndex(renderer->getMaxMotionFactor() + 11);
```
jwmelto: why are you bothering with `i` here? If you want the 9th element, just get the 9th element.

src/bzflag/DisplayMenu.cxx
blast007: No display options were removed.  They were just re-ordered, and two of the debugging options were merged into a single item.

src/bzflag/GUIOptionsMenu.cxx
blast007: The whole menu font size thing is a bit wonky.  We don't have any sort of automatic scaling depending on how many menu items are in a menu so we just tweak the number a bit to make it look okay.

src/bzflag/DisplayMenu.cxx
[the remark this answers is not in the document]
blast007: Yeah, that's outside the scope of this PR.  Our menu system as a whole needs a revamp or a replacement.

src/bzflag/DisplayMenu.cxx
blast007: Is there even a reason we are casting at all?  MainWindow::getWindow() already returns a BzfWindow*.

blast007:
I'm going to leave joystick changes for another time.

PR #244, opened 2020-02-09
Rework code that use glRect

atupone:
Reducing the number of place glRect is called

### DIFF
--- a/src/bzflag/ControlPanel.cxx
+++ b/src/bzflag/ControlPanel.cxx
@@ -360,17 +360,23 @@ void            ControlPanel::render(SceneRenderer& _renderer)
         if (BZDBCache::blend && _renderer.getPanelOpacity() < 1.0f)
             glEnable(GL_BLEND);
 
+        GLint x1 = messageAreaPixels[0];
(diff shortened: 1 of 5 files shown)


macsforme:
What is the point of this? It looks like you replaced one immediate-mode function (which at least made the code more concise) with another immediate-mode function which requires more lines of code to do the same thing.

atupone:
You are right, but it is just a step to do everything in retained mode. Maybe I could delegate the openGL calls to a class and do later the step to drawn in a more modern way,

macsforme:
I don't see this as helpful. There's nothing wrong with a shortcut function to draw a rectangle if we need to just draw a rectangle. When we're transitioning to a more modern OpenGL specification, or OpenGL ES which doesn't have the function, the simplest method (IMHO) is to write a stub function, like I did in my OpenGL ES port: https://github.com/macsforme/bzflag-embedded/blob/embedded/src/ogl/OpenGLESStubs.cxx#L102-L119

If and when we render something other than a simple rectangle there, or if we want to adapt it to a VBO so we're not streaming the vertex data every call, then it would make sense to me to do something like this. I just don't see how this is a particularly helpful transition.

PR #247, opened 2020-04-10
Optional replacement of short radar forward tick with dashed forward line

Paula90:
- Draw a dashed line in radar in tank heading direction instead of small tick at radar border
- Add GUI menu entry "Radar Forward Marker:" Tick/Line

... a big aiming aid for me ;-)

### DIFF
--- a/include/BZDBCache.h
+++ b/include/BZDBCache.h
@@ -65,6 +65,7 @@ public:
     static Int   radarStyle;
     static Float radarTankPixels;
     static Int   leadingShotLine;
+    static Int   radarForwardMarker;
     static Float linedRadarShots;
     static Float sizedRadarShots;
     static Int   radarPosition;
--- a/src/bzflag/GUIOptionsMenu.cxx
+++ b/src/bzflag/GUIOptionsMenu.cxx
@@ -132,6 +132,17 @@ GUIOptionsMenu::GUIOptionsMenu()
     option->update();
     listHUD.push_back(option);
 
+    // radar marker for forward direction
+    option = new HUDuiList;
+    option->setFontFace(fontFace);
+    option->setLabel("Radar Forward Marker:");
+    option->setCallback(callback, "m");
+    options = &option->getList();
+    options->push_back(std::string("Tick"));
+    options->push_back(std::string("Line"));
+    option->update();
+    listHUD.push_back(option);
+
     // set radar position
     option = new HUDuiList;
     option->setFontFace(fontFace);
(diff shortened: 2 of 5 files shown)


The-Noah:
It would be nice to have a screenshot so we can see what it looks like without having to download and compile it.

Paula90:
Sorry, my first contribution at all ... a little unexperienced ;-) 

bzflag_radar_line

The-Noah:
That's okay, I'm still new too :)
It might be just me but it looks slightly off-centered.

image

Paula90:
You're right, but isn't this natural? Given the center "square" (myTank) has an even pixel size, the dashed line cannot be exactly centered to this.
From the code, it's just a longer version of the existing tick ;-)

The-Noah:
Ah that's right 🙂

allejo:
Hmm... I'm not sure how I feel about this. For one, I would think this should be in a proto-break if this gets merged in; this feels like it's giving players with the most up to date client an unfair advantage in aiming. And on that note, I have my concerns that because it's providing aim assistance, that it'd be classified more like a cheat than a feature.

blast007:
There was a [patch][1] like this [accepted into the 2.1 development branch back in 2006][2], except that it was a solid line instead of a dashed line.  At a quick glance, I think this PR implements the drawing code better than the patch from 2006, but I'll have to take a closer look at it later.

However, I would also agree with @allejo that this should be added with a new major release of the game instead of a 2.4 compatible version.

[1]: https://sourceforge.net/p/bzflag/patches/363/
[2]: https://github.com/BZFlag-Dev/bzflag/commit/8e1531b2d90b1231b11c71e5b9d27827b419f767

macsforme:
According to the spec, `GL_LINE_STIPPLE` doesn't appear to be a valid argument to `glPushAttrib()`. Am I missing something?

Paula90:
> According to the spec, `GL_LINE_STIPPLE` doesn't appear to be a valid argument to `glPushAttrib()`. Am I missing something?

You're right, should be `GL_LINE_BIT` or `GL_ENABLE_BIT` right? Not sure from the spec whether there's a semantic difference in this case between both arguments.

macsforme:
The spec is a little confusing to me with respect to the terms "enable bit" versus "flag," but I would include both to be safe (e.g., `glPushAttrib(GL_ENABLE_BIT | GL_LINE_BIT)`). I don't see any downsides to using that approach in your proposed patch.

PR #249, opened 2020-04-21
Fix autogen having CRLF line endings

The-Noah:
This fixes `autogen.sh` having CRLF line endings (on Windows).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+autogen.sh text eol=lf


allejo:
I appreciate the PR for this, but if we're going to introduce a `.gitattributes` file, I'd much rather prefer it be a bit more fleshed out with more than a single file in it. e.g. This file won't define CRLF/LF standards on any of the hundreds of files that we have; `autogen.sh` isn't _that_ special 😄

Issue #246, opened 2020-03-21
Aggregated issues related to SDL 2 window creation


macsforme:
Here I attempted to summarize all the current information about our problems creating windows with SDL 2. This is mostly other people's work, so please edit or add to the info here if needed.

### BZFlag Issues Reported

#201 bzflag is not working properly with 3+ monitors

This is caused by a bug in SDL 2. If you have a primary display with a lower resolution than a secondary display and you launch BZFlag, the game window shows up on the secondary display (we always request the primary display), and SDL throws a resize event with the resolution of the secondary display. In addition to the window going onto the wrong display, this confuses BZFlag due to our window resize event handling. See #238 below for a workaround in progress, and see [SDL #4990](https://bugzilla.libsdl.org/show_bug.cgi?id=4990) for the report on the SDL bug tracker.

#241 Fullscreen toggling on macOS broken

When switching between a maximized window and a native fullscreen window via F1 on macOS, then pressing F1 again to switch back, the window is no longer maximized and the window content is offset by approximately the height of the window title bar. Our logic in `SDL2Window::create()` where we compare the resolutions before and after may be to blame. It may be worth comparing behaviors between the several methods available in SDL 2 to switch between windowed and fullscreen mode (e.g., `SDL_SetWindowFullscreen()` vs. `SDL_MaximizeWindow()`/`SDL_RestoreWindow()` vs. destroying and recreating the window).

#245 Retina scaling broken on macOS (again) in 10.15 Catalina

Based on a single report from a user that the window content is not properly laid out when using retina display scaling in macOS. We need to try to reproduce and confirm this issue.

#88 Resolution testing race condition

Not strictly related, but if the ultimate solution involves deeper adjustments to our fullscreen resolution picker, this one either might be worth fixing at the same time, or might go away on its own.

### SDL Bugs Reported

[SDL #4990](https://bugzilla.libsdl.org/show_bug.cgi?id=4990) SDL2 fullscreen windows on Linux are created on secondary monitor

See our #201 above.

~~[SDL #4177](https://bugzilla.libsdl.org/show_bug.cgi?id=4177) SDL_MinimizeWindow() fails to minimize fullscreen window on macOS~~

~~The SDL 2 team reports that this bug is fixed in their development code. After we verify this on our end, and after the fix makes it into the next SDL 2 release, we should remove our workaround code.~~ **Confirmed fixed.**

[...]

Issue #245, opened 2020-03-15
Retina scaling broken on macOS (again) in 10.15 Catalina


macsforme:
I asked for as many details as possible, but here is what we have so far:

> &lt;ashvala> on macOS 10.15, you can't have the full retina scaling 
> &lt;ashvala> for bz, i.e 
> &lt;ashvala> https://www.dropbox.com/s/y2zih1pkehbbf9o/Screen%20Shot%202020-03-14%20at%202.08.58%20AM.png?dl=0 
> &lt;ashvala> this is with the stock bz client. I've tried to compile it and it doesn't make a difference 
> &lt;ashvala> additionally, this is on AMD Radeon 560x. I've tried switching the GPU and it doesn't help
> &lt;ashvala> I have a feeling it might have something to do with macOS Catalina's increasingly awful support for OpenGL, but, I figure it's worth flagging with you

![Screen Shot 2020-03-14 at 2 08 58 AM](https://user-images.githubusercontent.com/4955448/76693241-2e98b300-661f-11ea-819d-e772b081529c.png)

TankUp-BZFlag:
I have the same Problem. This is the reason, why I switched back to Mojave three times. Had this issue with Catalina 15.1, 15.2 and 15.3! I have an iMac retina late 2015, Intel Iris Pro Graphics 6200.

macsforme:
> I have the same Problem. This is the reason, why I switched back to Mojave three times. Had this issue with Catalina 15.1, 15.2 and 15.3! I have an iMac retina late 2015, Intel Iris Pro Graphics 6200.

Any more details you can provide? For example, did the issue appear only on certain display scaling settings and not others? Did the issue exist when BZFlag was in windowed mode? The more details we have about this issue, the better our chances of fixing it, especially because AFAIK none of our developers have regular access to a Mac with a retina display at the moment.

TankUp-BZFlag:
Hi,

The full screen mode works perfectly with exactly half of the resolution of the retina display. Only that looks stupid. The higher I choose the resolution, the more the screen moves down. When I choose full native retina resolution, the image is moved down exactly half of the monitor (as in the screeenshot of this thread). If I switch to windowed mode, I can choose the full retina resolution, but due to the scaling of the window, the appearance suffers, especially from the radar, it all seems very thick. I could test beta versions or send log files, if needed. Only tell me what to do...

macsforme:
[...]

Issue #241, opened 2020-01-04
Fullscreen toggling on macOS broken


macsforme:
An issue with fullscreen toggling on macOS was introduced in 1fa9934d.

**Steps to replicate:**
1. Upon launch, switch to windowed mode via F1.
2. Maximize the window via the toolbar.
3. Press F1 to switch to regular fullscreen mode.
4. Press F1 again to switch back to windowed mode.
5. The window content is now offset and partially cut off at the top.

![Screen Shot 2020-01-04 at 1 33 31 AM](https://user-images.githubusercontent.com/4955448/71762873-7b7b4300-2e92-11ea-8fee-b5b76394a5de.png)

**Expected behavior:**

- BZFlag should return to standard windowed mode (non-maximized), but the drawing area should be resized to fit in the new window.

This was discovered while testing whether https://github.com/atupone/bzflag/blob/3mon/src/platform/SDL2Window.cxx#L268-L314 is still required or not. Incidentally, that event discard never seems to fire anymore, so we may be able to eliminate that section once we have a working window creation process again via #228 or #238. However, removing that section has no effect on this issue.

atupone:
I have seen this bug first on 2.4.8

Before that (2.4.6) the resultant window is the same as after toggling the window toolbar 
The 2.4.8 added the little offset on y

macsforme:
After pondering this further, I believe the issue is caused by the game switching into a standard (non-maximized) window upon exiting fullscreen, whereas it had a maximized window matching the fullscreen resolution prior to switching into fullscreen mode. Because a window with the fullscreen resolution does not fit into a standard window with a title bar and other OS GUI elements, the window contents become offset. Furthermore, our window creation mechanism does not currently have a way of telling the game that we got back a different resolution than we requested.

A solution may be to maximize the window again via the SDL 2 API after exiting fullscreen, if we are on macOS and detect that a window is being requested which matches the fullscreen resolution.

Issue #88, opened 2017-09-16
Resolution testing race condition


matheweis:
When testing a new screen resolution **R1**, if a user attempts to test _another_ resolution **R2** before the test of **R1** is complete, **R2** will be marked in the UI as "unavailable", and the user will no longer be able to select it for use.

PR #251, opened 2020-05-07
Fix multiple issues with SDL 2 window creation

macsforme:
Issue #246 lists several bugs with our SDL 2 window creation process. While digging into several of those, I also discovered several other regressions (changes that broke other features, code that was transplanted but not entirely, etc.) introduced into our window creation process over several commits. This fixes several issues with our window creation process, including #201 (this is basically a rework of #238) for the multiple monitor Linux issue, #241 for the fullscreen toggle issue on macOS, and some other issues I discovered along the way. ~~I still haven't figured out #245 yet.~~

### DIFF
--- a/src/platform/SDL2Window.cxx
+++ b/src/platform/SDL2Window.cxx
@@ -32,6 +32,9 @@ SDLWindow::~SDLWindow()
 {
     // Restore the original gamma when we exit the client
     setGamma(origGamma);
+
+    if (windowId != NULL)
+        SDL_DestroyWindow(windowId);
 }
 
 void SDLWindow::setTitle(const char *_title)
--- a/src/platform/SDL2Window.cxx
+++ b/src/platform/SDL2Window.cxx
@@ -107,7 +107,7 @@ void SDLWindow::setSize(int _width, int _height)
 void SDLWindow::getSize(int& width, int& height) const
 {
     if (fullScreen)
-        const_cast<SDLDisplay *>((const SDLDisplay *)getDisplay())->getWindowSize(width, height);
+        const_cast<SDLDisplay *>(static_cast<const SDLDisplay *>(getDisplay()))->getWindowSize(width, height);
     else
     {
         width  = base_width;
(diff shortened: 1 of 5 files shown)


macsforme:
#245 was fixed by updating to the latest SDL 2 version, so scratching that from the above list. So this should now fix all the issues listed in #246 (other than #88, which doesn't necessarily need to be addressed contemporaneously with these other issues).

blast007:
On Linux, it does correctly start the game on the primary monitor when the primary monitor has a lower resolution (so, the issue in #201). However, with either a single or multiple displays, changing the resolution in-game triggers an infinite loop of the screen going blank and returning to the desktop. (With my laptop set to only display on the external display, it just went into power saving mode when I changed resolution. Might be repeatedly changing resolutions and just confused that monitor.)

blast007:
@macsforme mentioned to me that the resolution change infinite loop on Linux is happening in the 2.4.20 release, so it seems that may be a regression there.  I'll be checking into that tomorrow.

blast007:
I tested older BZFlag versions with SDL 2.0.9, back to 2.4.8, and they're also showing the infinite loop issue, so I guess it's not a recent regression, at least with our code.  I will do some further testing.

macsforme:
Trying to summarize the remaining issues that have been observed:

  - On Linux, in certain conditions, multiple window flashes are seen when creating the window or changing resolution
  - On Linux, the original desktop resolution is not restored upon exit after the game has executed a modeset
  - On Linux, in certain conditions, pressing F4 does not properly iconify the window
  - On Linux, using certain window managers, resizing the window by dragging the corner triggers odd behavior, as seen [here](https://youtu.be/Jk3tNsb5k1Y)

Additionally, we did mark #245 resolved after updating SDL seemed to fix the issue, but there was a comment from the original reporting user that there are still some issues with retina scaling. I do not know of any other complaints, and I do not have the needed hardware to test further. If someone wants to try running the game on macOS Catalina under several different retina scaling settings just so we can do our due diligence, I would consider that matter fully addressed for the time being.

src/platform/SDL2Window.cxx
```diff
@@ -32,6 +32,9 @@ SDLWindow::~SDLWindow()
 {
     // Restore the original gamma when we exit the client
     setGamma(origGamma);
+
+    if (windowId != NULL)
+        SDL_DestroyWindow(windowId);
```
jwmelto, line 37: if you destroy the window, shouldn't you set it to null? (or better, `nullptr`)

src/platform/SDL2Window.cxx
```diff
@@ -148,7 +151,10 @@ void SDLWindow::swapBuffers()
     if (! SDL_GL_GetSwapInterval())
         return;
 
-    const int maxRunawayFPS = 65;
+    int maxRunawayFPS = 65;
+    SDL_DisplayMode desktopDisplayMode;
+    if(SDL_GetDesktopDisplayMode(0, &desktopDisplayMode) == 0)
```
jwmelto, line 156:
```suggestion
    if (SDL_GetDesktopDisplayMode(0, &desktopDisplayMode) == 0)
```

src/platform/SDL2Window.cxx
```diff
@@ -165,6 +171,24 @@ void SDLWindow::swapBuffers()
 #endif //__APPLE__
 }
 
+// For some reason, when creating a new fullscreen window on Linux with a different resolution than before, SDL throws
+// a resize event with the old window resolution, which is not what we want. This function is called to filter SDL
+// window resize events right after the resolution change and adjust the resolution to the correct one.
+#ifdef __linux__
```
jwmelto, line 177: does it **need** to be restricted to Linux?

src/platform/SDL2Window.cxx
```diff
@@ -259,7 +283,32 @@ bool SDLWindow::create(void)
             return false;
         }
     }
-#endif
+
+    // Depending on the distribution (or possibly the window manager), SDL will not recognize the new window resolution
+    // and will keep returning the previous resolution when SDL_GetWindowSize() is called, until a period of time has
+    // passed and events have been pumped. Wait up to two seconds for the correct resolution to start being returned,
+    // checking every quarter second, to avoid repeating the window destruction/re-creation process based on bad data.
```
jwmelto, line 290: I wonder if there is a way to flush the queue? This seems like an exceptionally awkward work-around. Or maybe we don't understand the resize event logic?

src/platform/SDL2Window.cxx
macsforme: Agreed. I'll make a mental note to fix this when I do a final interactive rebase to fix and clean up a few things prior to merging.

src/platform/SDL2Window.cxx
macsforme: Not sure. In general, I would like to keep platform-specific workarounds within `#ifdef` guards, because we have a tendency to break other platforms while fixing an issue on one of them. Keeping the unknowns to a minimum is a plus in my opinion.

src/platform/SDL2Window.cxx
macsforme: Completely agreed that the workaround is awkward. `SDL_PumpEvents()` *is* flushing the queue in this case. After extensive testing, it appeared to me that *both* a time delay *and* pumping events are required in order to reach the point where the correct window resolution is returned. My best guess is that the window creation process in SDL returns before the modeset is complete. It took extensive testing to get to this point. However, if anyone else has a better solution, I'd be all for it.

macsforme:
The following additional issues have now been resolved:

* Multiple window flashes when starting the game on Linux
* The desktop resolution was not restored on exit on Linux after using a scaled fullscreen resolution (thanks @blast007)
* Issues with iconify and application switching on Linux
* [Glitching](https://youtu.be/Jk3tNsb5k1Y) when resizing the window by dragging the corner on Linux
* Several other previously undiscovered issues, primarily related to using a scaled fullscreen resolution

Some of these fixes (specifically the F4/application switching fixes on Linux) required changes to logic outside of the platform module.  Accordingly, this could use some extensive testing, especially related to launching in windowed versus fullscreen mode, using scaled versus native fullscreen resolutions, changing resolution, application switching, iconifying, window resizing, and various combinations/orders of the above, with various monitor counts and arrangements.  I did hammer on macOS and Linux (Ubuntu) pretty hard, but Windows especially still needs to be checked.

I also want to do one final interactive rebase to squash a few commits together and fix a few other things. Other than that, if no further issues are discovered, I think this is good to merge.

blast007:
I've finished my testing on Windows 7 (and a bit on 10), Linux (Xorg and Wayland), and macOS and I have found no regressions and it has fixed the majority of issues with our SDL2 code.  I agree that this is ready to merge.

macsforme:
> Additionally, we did mark #245 resolved after updating SDL seemed to fix the issue, but there was a comment from the original reporting user that there are still some issues with retina scaling. I do not know of any other complaints, and I do not have the needed hardware to test further. If someone wants to try running the game on macOS Catalina under several different retina scaling settings just so we can do our due diligence, I would consider that matter fully addressed for the time being.

I managed to borrow a 15" Retina MacBook Pro running Catalina.  After building this PR, I tried every display scaling setting, and within those I tried launching the game with and without a resolution set in the configuration.  Everything appeared to work fine regardless of the scaling setting.  I did observe that the default resolution in the game would scale down with the more zoomed in scaling settings, and at the standard scaling setting the default resolution was half (so not high-DPI, but rather matching the "logical" resolution).  To me, this is the expected and desired behavior, and users can always manually set a higher resolution if they want that.  I will consider this concern resolved.

@ashvala, feel free to comment if you run into any more issues once this is merged.

PR #252, opened 2020-05-07
Add GitHub workflow for checking formatting

allejo:
Here are two example runs:

- Failed: https://github.com/allejo/bzflag/actions/runs/97748415
- Successful: https://github.com/allejo/bzflag/actions/runs/97749302

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,14 @@
+name: Formatting
+on: [push]
+jobs:
+  astyle:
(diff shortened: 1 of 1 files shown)


The-Noah:
Forgive me if I'm missing something, but to me it looks like they both pass.

allejo:
> Forgive me if I'm missing something, but to me it looks like they both pass.

Ah good catch! Fixed the links

PR #253, opened 2020-05-15
Better gcc compiler version check. Previous check failed for gcc-10

atupone:


atupone:
> Can you please describe the motivation for this change?
Previous check was "10.1.0" >= "4.3" and it did not succeed.
the only comparison could be made as string as 10.1.0 is not a number
the change is to make each version pieces as a 2 digit number and concatenate all
so 10.1.0 become 100100
and 8.3.1 become 080101
and also 4.3 to become 040300

I understand that some older gcc version gave us only two version pieces: major.minor

configure.ac
```diff
@@ -744,12 +744,18 @@ case $host_os in
 	CONF_CXXFLAGS="$CONF_CXXFLAGS $FLAGS"
 	case $host_vendor in
 	     pc)
-	       if expr `$CC -dumpversion` \>= "3.4" > /dev/null ; then
+               # Get gcc version as a triplet major.minor.patch or for some case major.minor
+               # if minor or patch are 2 digits remove the '.'
+               # if minor or patch are 1 digits remove the '.' and add a 0
+               # if there is no patch part, so the resulting version is 3 or 4 digits only add 00 ath the end
+               # This is to change 4.7.1 in 040701 or 3.4 in 030400
+               GCCVER=`$CC -dumpversion | sed -e 's|\.\([[0-9]][[0-9]]\)|\1|g' -e 's|\.\([[0-9]]\)|0\1|g' -e 's|^[[0-9]]\{3,4\}$|&00|'`
```
jwmelto, line 752: This doesn't work for MacOS 10.13 (should it? does it need to?
```
Goliath:~ jwmelto$ gcc -dumpversion | sed -e 's|\.\([[0-9]][[0-9]]\)|\1|g' -e 's|\.\([[0-9]]\)|0\1|g' -e 's|^[[0-9]]\{3,4\}$|&00|'
4.2.1
```
```suggestion
               GCCVER=`gcc -dumpversion | awk -F. '{printf "%02d%02d%02d", $1,$2,$3}'`
```
or use [ax_compare_version](https://www.gnu.org/software/autoconf-archive/ax_compare_version.html) as @blast007 suggested

jwmelto:
That looks familiar. Good change.

jwmelto:
Now I know why this looks familiar. See #236

Issue #72, opened 2017-03-09
Joystick axis range


macsforme:
The raw joystick axis readings are currently subject to several obscure calculations prior to being used for driving speed/rotation. One of the factors is based on the mousebox size. While this does allow for some control over the joystick sensitivity, the results do not seem ideal. For example, at mousebox size -5, my gamepad stick reaches the full value at approximately 10% in one axis, and about 15% in the other. This makes it very difficult to play the game with a joystick. The obvious solution is to just map the actual values onto a -1, 1 scale. However, with a circular axis (as some joysticks have), with the stick in a diagonal direction, the maximum in each axis will be (√2)/2 (or approximately 0.7), putting joystick players at a disadvantage. While a calculation is possible to map a circular area onto a square, some sticks (such as mine) do not have a 1.0 magnitude at all angles. Some users may also need/want to adjust the sensitivity of their sticks.

It appears the solution will be to map circular sticks to the control box (square), to have some kind of test to determine the full range of the joystick, then mapping it appropriately and also offering the user a way to adjust the sensitivity of the joystick. Some kind of visual feedback will probably also be helpful.

macsforme:
Looking at this again, I think the solution is to detach the joystick sensitivity from the mousebox entirely and just use the raw axis values, and then having the following configurable settings in the menu:

- Assign joystick axes (including negated axes, e.g., 1, -1, 2, -2, etc.) to
  - Movement (forward/backward)
  - Rotation (left/right)
- Set sensitivity (fractional part of axis to be actually used for control)
- Set dead zone (maybe we can find a sane default for this? but then each controller is different...)
- Map circular zone to square (true/false, so that holding a stick at 45 degrees will read 1,1 rather than ~0.7,~0.7)
- Ramp type (linear ramp, exponential ramp)

I think that about covers every contingency. Thoughts?

PR #255, opened 2020-06-21
Improved joystick support

macsforme:
This PR implements improved joystick handling per my comments [here on issue #72](https://github.com/BZFlag-Dev/bzflag/issues/72#issuecomment-549397191).

[This video](https://youtu.be/5pwImEa5vf8) explains the changes better than I could by typing it out here, so I will leave it at that.

I tested this on my SteelSeries Nimbus, but it would help if anyone with a different model of joystick or controller could also give it a whirl and make sure the effects of the joystick input settings are as expected.


Ratfink:
The ramp options in the menu aren't named quite right, since they're really using a polynomial function, not exponential.  I'd suggest being more concise and just calling the three options "Linear," "Quadratic," and "Cubic."

src/platform/DXJoystick.cxx
```diff
@@ -255,33 +255,29 @@ bool          DXJoystick::joystick() const
...
-    else if (yAxis == "Rx") y = state.lRx;
-    else if (yAxis == "Ry") y = state.lRy;
-    else if (yAxis == "Rz") y = state.lRz;
-    else if (yAxis == "Slider 1") y = state.rglSlider[0];
-    else if (yAxis == "Slider 2") y = state.rglSlider[1];
-
-    // ballistics
-    x = (x * abs(x)) / 1000;
-    y = (y * abs(y)) / 1000;
+    if (xAxis == "X")  x = float(state.lX) / 1000.0f;
```
jwmelto, line 264: lots of superfluous casting here (but at least they are function-style casts and not C-style casts)... the divide by float will do the type promotion, and the type of `x` will narrow as needed.

jwmelto (approved):
I wonder if we want to persist in 32-bit floating point interfaces/calculations ... all modern hardware (except for GPU) is faster in 64-bit operations. What are we gaining by constraining precision?

src/platform/DXJoystick.cxx
macsforme: Fixed... thanks.

macsforme:
This project kind of stalled after I learned that one of the most common controllers out there (the Xbox One controller) has triggers which are treated as axes rather than buttons.  Our code currently has no support for the concept of an axis triggering a button event, so this makes the triggers on the Xbox One controller basically useless for their intended purpose.  Implementing support for an axis to trigger a button would require consideration of actuation point, release point, etc.  Expanding the underlying joystick support code is complicated by the number of platform implementations we have on top of it (EvDev, USB, Windows, SDL 1, SDL 2).  Development time and available hardware for testing are also scarce resources.

Any thoughts on how to move forward would be welcome.  One valid option may be to simply merge as-is, and ignore the lack of axis->button support for the time being, since we will be no worse off than we were before this patch.  **However**, the EvDev/USB/Windows joystick code should be updated to use the new floating-point axis readings versus the old integer readings (I believe Windows is done, but may need testing), if any of that code is even connected to anything anymore.

blast007:
We should remove the non-SDL2 joystick code. That was my plan at least in master once we're done worrying about merging in 2.4 changes. The main reason the Windows joystick code is still in use is because of its Force Feedback support. I do have the start of code that added haptic feedback to our SDL2 joystick code, though.

I was considering an analog-to-digital conversion already to handle menu/UI navigation with analog sticks.

We could also look into use the [GameController API](https://wiki.libsdl.org/CategoryGameController) that SDL2 provides which has mappings for [various controllers](https://raw.githubusercontent.com/gabomdq/SDL_GameControllerDB/master/gamecontrollerdb.txt).

macsforme:
That last commit ([dcc3d87](https://github.com/BZFlag-Dev/bzflag/pull/255/commits/dcc3d87b0a589b88c2d85961405b8ec374581d66)) fixes my last concern with this PR, which was an issue where the real coordinates and effective coordinates did not always align due to incorrect math. [This video](https://youtu.be/Seot2qzjORs) shows that the two are now properly aligned (the red/green lines were a temporary visual aid and are not present in the main PR). Due to the math change, a second set of eyes on that commit would be welcome.

With that, I believe this is ready to merge assuming no other outstanding concerns.

PR #259, opened 2020-08-07
Buffer udp

atupone:
The current status:
Any time there is a "begin shot" there is also an update of the tank position.
The idea about sending these two messages was to see bullets starting where a tank is.
Unfortunately these packets are sent always using 2 different datagrams, at least on UDP, bzfs does not support UDP buffering.
When tank sends a stream of shots it also send a stream of player updates.

This pull request:
Make bzfs accept buffered udp datagram (1st commit)

Make client merge player update and begin shot on the same datagram (2nd commit)

Client build with the second commit cannot shoot anymore (in UDP) with old server.

New server accept new and old clients

My proposal is to add the first commit, the bzfs handling of udp buffered messages, to 2.4, and the client changes to leave for the next protocol break. In the mean time there is a possibility to test if this reduce the massive FramePerSecond drop that sometime happens.

### DIFF
--- a/src/bzfs/bzfs.cxx
+++ b/src/bzfs/bzfs.cxx
@@ -7665,7 +7665,7 @@ int main(int argc, char **argv)
                     unsigned char ubuf[MaxPacketLen];
                     bool     udpLinkRequest;
                     // interface to the UDP Receive routines
-                    int buffLen;
+                    int buffLen = MaxPacketLen;
                     int id = NetHandler::udpReceive((char *) ubuf, buffLen, &uaddr,
                                                     udpLinkRequest);
                     if (id == -1)
@@ -7688,12 +7688,12 @@ int main(int argc, char **argv)
                             sendUDPupdate(id);
 
                         unsigned char *buf = ubuf;
-                        uint16_t len;
 
                         // handle the command for UDP
                         // checking if more messages are on the same datagram
                         while (buffLen >= 4)
                         {
+                            uint16_t len;
                             handleCommand(id, buf, true);
                             nboUnpackUShort(buf, len);
                             buf     += len + 4;
--- a/src/bzfs/bzfs.cxx
+++ b/src/bzfs/bzfs.cxx
@@ -7665,7 +7665,8 @@ int main(int argc, char **argv)
                     unsigned char ubuf[MaxPacketLen];
                     bool     udpLinkRequest;
                     // interface to the UDP Receive routines
-                    int id = NetHandler::udpReceive((char *) ubuf, &uaddr,
+                    int buffLen;
+                    int id = NetHandler::udpReceive((char *) ubuf, buffLen, &uaddr,
                                                     udpLinkRequest);
                     if (id == -1)
                         break;
@@ -7686,8 +7687,18 @@ int main(int argc, char **argv)
                             // send client the message that we are ready for him
                             sendUDPupdate(id);
 
+                        unsigned char *buf = ubuf;
+                        uint16_t len;
+
                         // handle the command for UDP
-                        handleCommand(id, ubuf, true);
+                        // checking if more messages are on the same datagram
+                        while (buffLen >= 4)
+                        {
+                            handleCommand(id, buf, true);
+                            nboUnpackUShort(buf, len);
+                            buf     += len + 4;
+                            buffLen -= len + 4;
+                        }
 
                         // don't spend more than 250ms receiving udp
                         if (TimeKeeper::getCurrent() - receiveTime > 0.25f)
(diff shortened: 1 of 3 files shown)


src/bzflag/ServerLink.h
```diff
@@ -63,6 +63,7 @@ class ServerLink
     const PlayerId& getId() const;
     const char*     getVersion() const;
 
+    void        send(const void* msg, int len, bool udp);
```
jwmelto, line 66: I would not have this method either public, or an overload. The other methods in this class are based around a message id (`code`) whereas this sends a raw buffer. Given the constraints on the API, allowing an arbitrary blob to be sent seems like a bad idea.

I might name this something like `sendBuffer`.

You might also encapsulate the branches of this method into `sendUdpBuffer` and `sendTcpBuffer`, but that's subjective

src/bzfs/bzfs.cxx
```diff
@@ -7666,8 +7667,18 @@ int main(int argc, char **argv)
                             // send client the message that we are ready for him
                             sendUDPupdate(id);
 
+                        unsigned char *buf = ubuf;
+                        uint16_t len;
```
jwmelto, line 7671: this should be declared before use, inside the loop. Declaring it here (and not initializing it) increases complexity.

src/bzfs/bzfs.cxx
```diff
@@ -7645,7 +7645,8 @@ int main(int argc, char **argv)
                     unsigned char ubuf[MaxPacketLen];
                     bool     udpLinkRequest;
                     // interface to the UDP Receive routines
-                    int id = NetHandler::udpReceive((char *) ubuf, &uaddr,
+                    int buffLen;
```
jwmelto, line 7648: an unpleasant aspect of by-reference return values in functions is that it encourages this style of uninitialized declaration. Sure, the value will be set in the next call, but the next reader of this code might not know that. At a minimum, I would initialize all variables to reasonable defaults.

jwmelto (changes requested):
Given that this PR is in two parts, one of which breaks existing clients, I don't see how we can merge it as is.

Is there an observable problem that this is addressing? Is it simply a "good idea" optimization? There is certainly other improvement that could be made to the networking code; is this a band-aid on a severed artery, or is there a specific issue this will improve? Otherwise, a more extensive refactoring might be appropriate.

atupone:
> Given that this PR is in two parts, one of which breaks existing clients, I don't see how we can merge it as is.
> 
> Is there an observable problem that this is addressing? Is it simply a "good idea" optimization? There is certainly other improvement that could be made to the networking code; is this a band-aid on a severed artery, or is there a specific issue this will improve? Otherwise, a more extensive refactoring might be appropriate.

IIRC the PR applied to the server does not break existing client. It only allow buffered UDP to be handled.
The PR applied to client breaks on old server.

The optimization is to reduce the traffic on shoot messages. in the current code they are always preceded by tank position so we send two very short messages one after the other, This change will save for any shot, a IP and a UDP header to only send few bytes. More over, it can happen that in case of lost messages, you can loose the tank position, so can you see the shot starting from nowhere.
My proposal is to add this capability to the server code and after a while adding the client capability to the code

jwmelto:
My point is, the PR is applied to the **code base** (both server and client). This is a request to merge the submitted files (which include the client). As such, and since you point out that it breaks the client, it is not appropriate prior to a protocol change. In effect, it **is** a protocol change.

I'm not questioning whether reducing the traffic is a good idea; I'm questioning whether it will have a noticeable impact on game play (I truly don't know). Otherwise, this particular change falls into the category of premature optimization. 

Now, the nice thing about open source projects is that the time you spend is yours to spend as you wish, so premature optimization isn't as costly to the "organization" as when people are paid. I'm also not saying this change is a bad idea. I just want to know if the cost (in this case, protocol breakage) is worth the gain (unspecified).

macsforme:
Why is this going into 2.4 at all?  The more we add to 2.4, the more complex it becomes to merge these changes into master.  Furthermore, what play testing, simulations, etc. have been done (if any) on these changes?  If you want to make changes that affect our network logic, simulation, or other core parts of the game, especially this deep into a maintenance cycle on this branch, then they call for some testing/validation.

I am concerned about changes like this (your GM math changes, for another example) going into any branch without any mention of how they were tested and validated.  We've been through this before, when we tried to get 3.0 out, and we found that 2.99 had a lot of issues with the simulation and with things generally not working, which I perceive was because of widespread inattention to making sure stuff worked (including possible side effects) before committing it.  Considering several examples of a lot of tedious effort recently to fix issues, many of which should not have happened to begin with, I believe we need to get everyone on board with expectations of a more cautious and thorough approach up front to avoid issues happening down the road.

PR #261, opened 2020-08-26
Implement RogueGenocide as options

Zehra:
Compile tested patch which should implement "RogueGenocide" as an option within the "Genocide" plug-in.


allejo:
I'm against using arbitrary numbers for configuration options. I'd prefer having understandable flags like `allowRogue` and `allowSuicide` to be comma-separated as the plug-in's configuration.

```
-loadplugin genocide,allowRogue,allowSuicide
```

This way, we can also allow suicide genocide but not rogue genocide.

Zehra:
Would option for "suicide" apply to "teams" as well?
Since this could allow for where selfkills would not effect teams, which where selfkills trigger the effect in both 2.4.x and 2.x.
(However there was a patch provided for 2.99 which would have prevented selfkills from affecting teams, if the "**-noTeamKills**" option were enabled.)

Improved flags will be added next commit, just want to clarify above, as it would introduce changes to game play.

allejo:
Blegh I don't want to have so many different configuration options of Genocide. How about these options:

- `rogueAsTeam`
- `disableSuicide` (for everyone)

The default behavior will be to allow suicide ONLY for colored teams. Suicide for rogues will only automatically happen when `rogueAsTeam` is used. When `disableSuicide` is used, it will disable suicide for all teams including rogues.

Thoughts @blast007?

macsforme:
A complete implementation would also remove the RogueGenocide plugin from the source as well as (at least) from the makefile system. This change also calls for documentation in the ChangeLog.

Zehra:
Will update with both recommendations once consensus on game play is established.
My familiarity with Git is somewhat limited, so possibly a bit of time before next update.
(This way both actions may be done in a single go vs multiple commits for a single action.)

Zehra:
Ah, just realized what happened:

Currently with RogueGenocide and Genocide as plug-ins, the following states are possible:
1. Genocide is loaded by itself.
2. Genocide and RogueGenocide is loaded.
3. Genocide and RogueGenocide is loaded with "suicide" option.
No other states are possible:
(Since simply loading RogueGenocide by itself will do nothing.)

Due to this, it would be impossible to have "suicide" enabled, as nothing would occur.(Unless RogueGeno is enabled.)
(In the current version, hence my confusion of if an additional state would be affected, as self-kills triggering geno already effects teams and is not possible to disable it, either with options or "-noTeamKills" setting.)

For a future release, there was a patch which disabled genocide from effecting teams by self-kills, if "-noTeamKills" were enabled.
(However this may be an outdated goal.)

So in theory, there's only "2" options, with one option being a side option of a side option. (With current PR.)
Genocide being default, RogueGenocide being side option of Genocide, with "suicide" being an option of RogueGenocide.

---

Proposed modification results in the following possible states:
1. Genocide default.
2. Genocide with RogueGeno.
3. Genocide with RogueGeno and disable suicide.
4. Genocide with disable suicide.

The only changes which result from it are the following with regards to current and proposed game play is:
_4. Genocide with disable suicide._
This partially implements behavior of an accepted patch for a proposed version of the game and does not in any existing versions.
This change would mainly emulate said patch and be the only differing version of the game play.
(Except in the case of where team kills are enabled and "disable suicide" is enabled" is the main difference from patch.)
Hence proposed modification would only introduce a single game play modification.

plugins/genocide/genocide.cpp
```diff
@@ -16,29 +16,47 @@ class genocide : public bz_Plugin
...
 };
 
 BZ_PLUGIN(genocide)
 
-void genocide::Init ( const char* commandLine )
+void genocide::Init ( const char* cmdLine )
 {
     bz_debugMessage(4, "genocide plugin loaded");
-    if (commandLine)
+    if (strlen(cmdLine) != 0)
```
jwmelto, line 28: This is a fundamental change in functionality/stability.

One of the dangers of pointers is that they have the potential to be null. The previous code tested for null. Here you assume it's valid, and points to memory with a single null byte. 

Testing for an empty string is pretty pointless.

plugins/genocide/genocide.cpp
```diff
@@ -16,29 +16,47 @@ class genocide : public bz_Plugin
...
     {
-        int conf = atoi(commandLine);
-        if (conf == 1 || conf == 2)
-            rogueGenocide = true;
-        if (conf == 2)
-            allowSuicide = true;
+        if (strcmp(cmdLine, "rogueAsTeam,disableSuicide") == 0)
```
jwmelto, line 30: this class of methods are dangerous. Prefer `strncmp` and friends instead. Or, better yet, use `std::string` instead.

plugins/genocide/genocide.cpp
```diff
@@ -16,29 +16,47 @@ class genocide : public bz_Plugin
...
+            rogueAsTeam = true;
+        }
+        else if  (strcmp(cmdLine, "disableSuicide") == 0)
+        {
+            disableSuicide = true;
+        }
+        else
+        {
+            bz_debugMessage(0, "Error: Check arguments.");
+        }
```
jwmelto, line 51: this block could be simplified with `std::string::find`, as in:

```c++
auto cmd = std::string(cmdLine);
disableSuicide =  (cmd.find("disableSuicide") != cmd.npos);
rogueAsTeam    =  (cmd.find("rogueAsTeam")    != cmd.npos);
```

plugins/genocide/genocide.cpp
```diff
@@ -16,29 +16,47 @@ class genocide : public bz_Plugin
...
+        
     }
 
-    if (rogueGenocide == true)
-        bz_debugMessage(4, "rogue genocide enabled");
+    if (rogueAsTeam == true)
```
jwmelto, line 55: just a personal preference, but testing boolean variables against explicit boolean values is redundant.

Well named variables should be self-explanatory, as in
```c++
if (rogueAsTeam)
   ...
```

jwmelto (approved):

jwmelto:
by the way, don't just remove from the build; delete the files as well

Zehra:
Updated with suggestions/recommendations.
I'm not sure, but [I think I deleted the files as well](https://github.com/BZFlag-Dev/bzflag/pull/261/commits/3a61d52c8b1868bbbbd2957a2b10df51d0425b06)? (I might have missed something though.)

plugins/genocide/genocide.cpp
```diff
@@ -12,25 +12,37 @@ class genocide : public bz_Plugin
     {
         return "Genocide";
     }
-    virtual void Init ( const char* c );
+    virtual void Init ( const char* config );
     virtual void Cleanup ( void );
     virtual void Event ( bz_EventData *eventData );
+
+    bool disableSuicide = false;
```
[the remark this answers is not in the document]
Zehra: Within the _RogueGenocide_ plugin, it was done as such.
Any recommendations? Since I've seen both **private** and **protected** used in a few other plugins
For instance, _playHistoryTracker_ uses **protected**, _regFlag_ uses **private** and _airspawn_ uses **public.**

plugins/genocide/genocide.cpp
```diff
@@ -53,64 +66,67 @@ void genocide::Event(bz_EventData *eventData)
...
+                    bz_sendTextMessage(BZ_SERVER, targetID, "You were a victim of Rogue Genocide");
 
+                    // oops, I ended up killing myself (directly or indirectly) with Genocide!
+                    if (targetID == dieData->killerID)
+                        bz_sendTextMessage(BZ_SERVER, targetID, "You should be more careful with Genocide!");
+                }
+
+                bz_freePlayerRecord(playRec);
+            }
+        }
```
allejo, line 111: - These are identical loops, there's no need to split them for rogues only; combine the two.
- Is there a reason `eGenocideEffect` doesn't work for rogue tanks? Skimming through the source code, there doesn't seem to be anything special server-side or client-side disabling `eGenocideEffect` from affecting rogues. Can you test this behavior to make sure `eGenocideEffect` works or doesn't work?
- The "be careful message" shouldn't just apply to rogue genocide, that should apply to genocide in general, but should only appear if you're **not** rogue.

allejo (changes requested):

Issue #263, opened 2020-09-15
plugin_utils lib size seems too large


allejo:
For some reason, our `libplugin_utils.a` is coming out to be about 2M? Is that right? If so, where is the extra size coming from?

For example with SAMPLE_PLUGIN, when I link to plugin_utils I get a `.so` that's ~911K when it's not being linked, it comes out to ~35K. Shouldn't the compiler be optimizing out a lot more of that code?

jwmelto:
`.so` is fundamentally different from `.a`. It's generally not a good idea to link `.a` libraries into `.so` libraries. Shared objects need to have position-indepdent code (-fPIC or -kPIC or some variant, depending on the compiler); `.a` objects don't routinely have this enabled.

Now, if the size of `libplugin_utils.a` is a concern, we might consider changing it to a `.so` so that one instance in memory can service all plugins (smaller runtime footprint)

PR #264, opened 2020-09-16
 Reset CFLAGS and CXXFLAGS to a previous value to fix the large release builds.

blast007:
We noticed that release builds on Linux (and possibly other platforms that use autotools) were including debugging information in the binaries.  This change stops the -g compiler option from being automatically added to release builds.  Issue #263 reported part of the problem.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -794,10 +794,6 @@ fi
 if test x$enable_debug = xyes ; then
     CONF_CFLAGS="$CONF_CFLAGS -Werror -g -O0"
     CONF_CXXFLAGS="$CONF_CXXFLAGS -Werror -g -O0"
-    # revert automatic setting of CFLAGS and CXXFLAGS to prevent
-    # override of "-g -O0" put into CONF_CFLAGS and CONF_CXXFLAGS above
-    CFLAGS="$user_CFLAGS"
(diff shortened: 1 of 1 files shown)


jwmelto (approved):
There is definitely some magic going on here ... I've searched this entire file for "CFLAGS" and it's not possible to assert any knowledge of state.

On line 187, `user_CFLAGS` is set from `$ac_save_CFLAGS` (but that initialization is not seen)
On line 188, `user_CXXFLAGS` is set from `$CXXFLAGS` with a comment that `ac_save_CXXFLAGS` is not set yet. ???

`CONF_CFLAGS` gets appended to, in rational ways, as options are evaluated

Then, on (now) lines 816-820, there is this comment about reverting automatic setting... but they haven't been assigned? Neither have `user_CFLAGS` or `user_CXXFLAGS` been updated

All that is to say, I have to take your word for this. The whole thing looking painfully convoluted and difficult to maintain...

blast007:
I'll need to take a close look at the configure.ac file.  It do see now that the user_* options are only initialized when in debugging mode, but I could just always init those.  I'll poke around with it some more to see if I can figure out why it's adding -g in the first place.

Issue #268, opened 2020-10-21
Restoring minimized window on macOS 10.15 can result in an offset viewport


blast007:
On macOS 10.15, if you are running the game at a resolution different from the desktop resolution, restoring a minimized window by clicking on the application icon in the dock (instead of the minimized window icon) results in an offset viewport.  This occurs in both the stock 2.4.20 and with the code in PR 251.

blast007:
I need to test this a lot more to make sure it consistently works and doesn't have any side effects, but this seems to fix the offset viewport issue.  I added this to SDL2Display::setupEvent within the SDL_WINDOWEVENT switch statement.
```cpp
case SDL_WINDOWEVENT_MOVED:
{
    SDL_Window *window = SDL_GetWindowFromID(event.window.windowID);
    if (event.window.data2 < 0 && SDL_GetWindowFlags(window) & SDL_WINDOW_FULLSCREEN) {
        SDL_SetWindowFullscreen(window, 0);
        SDL_SetWindowFullscreen(window, SDL_WINDOW_FULLSCREEN);
    }
}
break;
```
So far I have only tested the change with a single display.  I'll need to test with multi-display configurations.

---

Here was a log of window events that led to the above change.  I was running this with a single display at 2560x1440 with the client resolution set to 1280x720.

Starting client full-screen at a non-desktop resolution:
```
2021-01-03 20:08:39.939 BZFlag[5334:69829] INFO: Window 1 shown
2021-01-03 20:08:39.940 BZFlag[5334:69829] INFO: Window 1 exposed
2021-01-03 20:08:39.940 BZFlag[5334:69829] INFO: Window 1 hidden
2021-01-03 20:08:39.940 BZFlag[5334:69829] INFO: Window 2 moved to 0,0
2021-01-03 20:08:39.940 BZFlag[5334:69829] INFO: Window 2 size changed to 1280x720
2021-01-03 20:08:39.940 BZFlag[5334:69829] INFO: Window 2 shown
2021-01-03 20:08:39.940 BZFlag[5334:69829] INFO: Window 2 gained keyboard focus
2021-01-03 20:08:39.940 BZFlag[5334:69829] INFO: Mouse entered window 2
2021-01-03 20:08:39.940 BZFlag[5334:69829] INFO: Window 2 exposed
```

Minimize client:
```
2021-01-03 20:08:52.439 BZFlag[5334:69829] INFO: Window 2 moved to 640,360
[...]

Issue #267, opened 2020-10-19
Saved resolution reset when quitting the game from taskbar/dock


blast007:
If you run the game at a resolution different from the desktop resolution, this saved resolution is lost if you quit the game from the Windows taskbar or the Linux/macOS dock.  In all cases, I was using SDL2.  It seems that the game is saving the current resolution when exiting, which overwrites the desired game resolution with whatever the desktop resolution of the primary display happens to be.  Note that this occurs both with the official 2.4.20 release and with the SDL2 fixes in PR 251.

Issue #266, opened 2020-10-19
macOS build confines mouse after being minimized


blast007:
When the mouse confinement mode is set to window, if you minimize the client with F4 (either while fullscreen or windowed), the mouse will continue to be confined to the previous game viewport and you have to "break free".  This is most noticeable when you're either running with dual displays (so that you're unable to move the mouse to the second display) or when running in a lower resolution or windowed mode.  This was tested on macOS 10.15.  Note that this occurs both with the official 2.4.20 release and with the SDL2 fixes in PR #251.

PR #269, opened 2020-11-08
More fixes for SDL2 window management

macsforme:
Several additional issues were discovered when validating the fixes for SDL2 window management in #251.  This can serve as a repository for fixes for these additional issues as they are completed.

- [x] #266 macOS build confines mouse after being minimized
- [ ] ~~#267 Saved resolution reset when quitting the game from taskbar/dock~~
- [x] #268 Restoring minimized window on macOS 10.15 can result in an offset viewport

### DIFF
--- a/src/bzflag/playing.cxx
+++ b/src/bzflag/playing.cxx
@@ -1035,7 +1035,7 @@ static void mouseClamp()
 {
     // only clamp when it might be useful
     if (HUDDialogStack::get()->isActive() || (myTank == NULL) || !myTank->isAlive() ||
-            myTank->isPaused() || (myTank->getTeam() == ObserverTeam))
+            myTank->isPaused() || (myTank->getTeam() == ObserverTeam) || unmapped)
     {
         mainWindow->disableConfineToMotionbox();
         return;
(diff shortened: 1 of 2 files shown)


blast007:
We'll ignore #267 for now.  That's not release critical and overall is a pretty minor bug.  We can worry about that one when/if we do further refactoring of our platform code.

blast007:
The changes in 766a58f6c0b45e2a209736c059d1869fc4781060 is causing Linux and macOS to not confine the mouse to the window.  I have a simpler fix that just disables confine to motion box if the window is unmapped and mouseClamp is true.  I'll test that solution a bit more.

blast007:
The alternative fix I have is as follows:
```diff
diff --git a/src/bzflag/playing.cxx b/src/bzflag/playing.cxx
index 7c3801d0e..f6850682b 100644
--- a/src/bzflag/playing.cxx
+++ b/src/bzflag/playing.cxx
@@ -1175,6 +1175,8 @@ static void     doEvent(BzfDisplay *disply)
 
         unmapped = true;
         mainWindow->ungrabMouse();
+        if (BZDB.isTrue("mouseClamp"))
+            mainWindow->disableConfineToMotionbox();
         break;
 
     case BzfEvent::KeyUp:
```

@macsforme Would you prefer I fix this PR to use this, or should I just directly apply that fix and 4735b2753a46f9c26f23266de4258bc01be8734e directly to our branch?

macsforme:
Let me test my fix again and compare it to yours.  This was a while ago, but I remember there being at least two parts to the problem (the motion box clamp, and the window clamp).

blast007:
I'm actually not sure if my alternative fix above actually fixes it.  I think when I was testing the issue just decided to stop happening.  I will need to test some more.

PR #285, opened 2021-07-07
Fix OBS Studio game capture on Windows.

blast007:
On Windows, trying to use game capture in OBS Studio would result in only the radar, chat panel, or altitude ticker areas redrawing.  This was due to our (incorrect?) use of glScissor.  This PR wraps our use of glScissor between glPushAttrib and glPopAttrib.  I also removed a few instances where we were setting the glScissor to the whole viewport, since that should no longer be necessary as we are restoring the scissor to the previous state after we are done.

I did notice that when the radar/panel opacity is set to full, the radar/panel area doesn't redraw in the OBS game capture.  I have not looked into that part yet.

### DIFF
--- a/src/bzflag/ControlPanel.cxx
+++ b/src/bzflag/ControlPanel.cxx
@@ -348,6 +348,8 @@ void            ControlPanel::render(SceneRenderer& _renderer)
     int   ay = (_renderer.getPanelOpacity() == 1.0f || !showTabs) ? 0
                : int(lineHeight + 4);
 
+    glPushAttrib(GL_SCISSOR_BIT);
+
     glScissor(x + messageAreaPixels[0] - 1,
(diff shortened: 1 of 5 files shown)


macsforme:
On macOS (Mojave), I'm getting this with the opacity set to full:

Screen Shot 2021-07-19 at 8 20 41 PM

The console does occasionally flash, maybe every few seconds or so.  On the stock 2.4.22 client, full opacity radar works as intended.

macsforme (approved):
With that fix for the MSAA framebuffer when radar opacity is full, I see no further regressions on macOS, Windows, or Linux, and OBS capture appears to work now in Windows.  There are a few new warnings when compiling using gcc on Linux that we should probably look at.  Other than that, this looks good to merge.  Nice work on this.

PR #289, opened 2021-10-03
Add Haiku build support

kenmays:
Adds OS build support for Haiku (https://www.haiku-os.org/).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -323,7 +323,7 @@ fi
 
 
 dnl detect a usable system regular expression library
-AC_CHECK_LIB(c, regcomp, LIBREGEX="-lc",
+AC_CHECK_LIB(root, regcomp, LIBREGEX="-lroot",
 [AC_CHECK_LIB(regex, regcomp, LIBREGEX="-lregex",
     [AC_CHECK_LIB(compat, regcomp, LIBREGEX="-lcompat", AC_MSG_ERROR([working regex library was not found]))]
 )]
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -323,7 +323,7 @@ fi
 
 
 dnl detect a usable system regular expression library
-AC_CHECK_LIB(c, regcomp, LIBREGEX="-lc",
+AC_CHECK_LIB(root, regcomp, LIBREGEX="-lroot",
```
jwmelto, line 326: I can't tell in context if this is a BeOS-specific portion or not, but I'm not sure it's a good idea to remove `-lc`; extending to check `-lroot` is fine.

configure.ac
```diff
@@ -363,7 +363,7 @@ AC_SUBST(LIBRT)
 AC_CHECK_LIB(m, sqrtf)
 
 # for BeOS - old network stack don't have those libs ( move it in the case switch ?)
-AC_CHECK_LIB([socket], [socket])
+AC_CHECK_LIB([network], [socket])
```
jwmelto, line 366: same here; I'm not at all literate in configure. Should this be an additional test instead of a change?

jwmelto (commented):
Wow. I remember being in the network code and seeing the BEOS defines and wondering if anyone would ever take that path. 

It's really cool to have you taking this on.

configure.ac
kenmays: Fixed.

configure.ac
kenmays: Fixed. Retest build. Works fine.

jwmelto (approved):
This looks fine to me

PR #291, opened 2021-10-25
GitHub action to ensure BZFlag builds

The-Noah:
This creates a GitHub Action that verifies BZFlag still compiles on push and in PRs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,9 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+    
+    - name: Update repositories
+      run: sudo apt-get update
 
     - name: Install dependencies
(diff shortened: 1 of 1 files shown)


blast007:
Our codeql-analysis.yml file already triggers a build on ubuntu-latest, though it currently does not build the client. This new action also installs SDL 1.2, which our 2.4 configure script won't use unless you explicitly tell it to use SDL1 via ```--with-SDL=1```. So currently this new action also won't build the client. In any case, SDL2 is our default and preferred version.

Overall, my thought on this is that if it's only going to be building for Linux, we should just fix up the CodeQL action so that it runs a complete build. I don't know if there's any advantage to running a CodeQL analysis on multiple platforms. If there isn't, then a separate action like this one that builds on macOS and Windows (and Linux?) would be nice.

blast007:
PR #317 from late 2022 has the CodeQL analysis run a full Linux build, including the client, so I'll close out this PR.  The only thing we might want to add, as I mentioned, would be Windows and macOS build tests as well.

PR #293, opened 2021-11-28
Reducing time spent opening/reading files when doing `/replay list ...` commands

### LINKED ISSUE #277: Extreme slowness with `/replay list` with 13,000 replays
So I'm kinda stumped on how to proceed with fixing this situation. On my Leagues United replay server, I have over 13,000 replays and running `/replay list` is extremely inefficient and leads to server freezes.

https://github.com/BZFlag-Dev/bzflag/blob/3be72c3aeba3f5f5e118a8279fb885333dc2b5b7/src/bzfs/RecordReplay.cxx#L865-L942

What I've gathered from reading through this code is: opening up 13,000 files and extracting the match date isn't the most efficient solution. Even though `sendFileList` will only output 100 results to the user as private messages, the server is still processing all 13,000 replays.

My thought for an immediate fix (2.4) is to change this functionality to use the filesystem "last modified" date for the sorting by time even though this can sometimes be inaccurate.

For 2.6, I would think having a replay server doing some sort of caching on boot to optimize these listings would be ideal.

red15:
Implemented:
* only read/open files that match the search query

Todo:
* limit number of results from `getFileList`

### DIFF
--- a/src/bzfs/RecordReplay.cxx
+++ b/src/bzfs/RecordReplay.cxx
@@ -162,7 +162,7 @@ static bool replaceWorldDatabase(ReplayHeader *h);
 static bool flagIsActive(FlagType *type);
 static bool packFlagTypes(char *flags, u32 *flagsSize);
 
-static bool getFileList(int playerIndex, std::vector<FileEntry>& entries);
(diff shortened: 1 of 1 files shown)


red15:
Linking to #277

PR #298, opened 2021-12-23
Use the audio buffer size the author really wants.

brad0:
Makes sound much smoother.

We have had this in in our OpenBSD port for almost 13 years.

### DIFF
--- a/src/platform/SDLMedia.cxx
+++ b/src/platform/SDLMedia.cxx
@@ -129,8 +129,7 @@ bool            SDLMedia::openAudio()
 
     // how big a fragment to use?  we want to hold at around 1/10th of
     // a second.
-    // probably SDL is using multiple buffering, make it a 3rd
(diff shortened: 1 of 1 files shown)


blast007:
In what situations does this make the audio smoother?  This adds a noticeable delay to audio.  We had [changed the multiplier from 0.8 to 0.3](https://github.com/BZFlag-Dev/bzflag/commit/5d135f8879b634145e12a8a2cf0967cc291976e3) back in 2004 to reduce the audio delay.

brad0:
Let me speak to our audio developer and see what he says.

brad0:
@ratchov

brad0:
@blast007 After further discussion and testing we have determined that although the patch was necessary 13 years ago things have changed with our kernel enough that it is no longer necessary. I am going to drop the PR and we will remove the one and only local patch.

PR #299, opened 2021-12-25
Add Support for Apple Silicon Macs

sanssecours:
Hi,

this minor update adds support for building BZFlag on Apple Silicon Macs. If I should change anything in the pull request then please just comment below. Thank you.

Kind regards,
   René

### DIFF
--- a/Xcode/BZFlag.xcodeproj/project.pbxproj
+++ b/Xcode/BZFlag.xcodeproj/project.pbxproj
@@ -7308,13 +7308,13 @@
 				HEADER_SEARCH_PATHS = (
 					../include,
 					../Xcode,
-					/usr/local/include,
 					/opt/homebrew/include,
+					/usr/local/include,
 				);
 				LD_RUNPATH_SEARCH_PATHS = "@executable_path/../Frameworks @executable_path/../PlugIns";
 				LIBRARY_SEARCH_PATHS = (
-					/usr/local/lib,
 					/opt/homebrew/lib,
+					/usr/local/lib,
 				);
 				MACOSX_DEPLOYMENT_TARGET = 10.7;
 				ONLY_ACTIVE_ARCH = YES;
@@ -7359,13 +7359,13 @@
 				HEADER_SEARCH_PATHS = (
 					../include,
 					../Xcode,
-					/usr/local/include,
 					/opt/homebrew/include,
+					/usr/local/include,
 				);
 				LD_RUNPATH_SEARCH_PATHS = "@executable_path/../Frameworks @executable_path/../PlugIns";
 				LIBRARY_SEARCH_PATHS = (
-					/usr/local/lib,
 					/opt/homebrew/lib,
+					/usr/local/lib,
 				);
 				MACOSX_DEPLOYMENT_TARGET = 10.7;
 				ONLY_ACTIVE_ARCH = NO;
--- a/Xcode/BZFlag.xcodeproj/project.pbxproj
+++ b/Xcode/BZFlag.xcodeproj/project.pbxproj
@@ -7309,9 +7309,13 @@
 					../include,
 					../Xcode,
 					/usr/local/include,
+					/opt/homebrew/include,
 				);
 				LD_RUNPATH_SEARCH_PATHS = "@executable_path/../Frameworks @executable_path/../PlugIns";
-				LIBRARY_SEARCH_PATHS = /usr/local/lib;
+				LIBRARY_SEARCH_PATHS = (
+					/usr/local/lib,
+					/opt/homebrew/lib,
+				);
 				MACOSX_DEPLOYMENT_TARGET = 10.7;
 				ONLY_ACTIVE_ARCH = YES;
 				OTHER_CPLUSPLUSFLAGS = (
@@ -7356,9 +7360,13 @@
 					../include,
 					../Xcode,
 					/usr/local/include,
+					/opt/homebrew/include,
 				);
 				LD_RUNPATH_SEARCH_PATHS = "@executable_path/../Frameworks @executable_path/../PlugIns";
-				LIBRARY_SEARCH_PATHS = /usr/local/lib;
+				LIBRARY_SEARCH_PATHS = (
+					/usr/local/lib,
+					/opt/homebrew/lib,
+				);
 				MACOSX_DEPLOYMENT_TARGET = 10.7;
 				ONLY_ACTIVE_ARCH = NO;
 				OTHER_CPLUSPLUSFLAGS = (
--- a/Xcode/BZFlag.xcodeproj/project.pbxproj
+++ b/Xcode/BZFlag.xcodeproj/project.pbxproj
@@ -7308,13 +7308,18 @@
 				HEADER_SEARCH_PATHS = (
 					../include,
 					../Xcode,
+					/usr/local/include,
 				);
-				"HEADER_SEARCH_PATHS[arch=*]" = /usr/local/include;
-				"HEADER_SEARCH_PATHS[arch=arm64]" = /opt/homebrew/include;
+				"HEADER_SEARCH_PATHS[arch=arm64]" = (
+					"$(inherited)",
+					/opt/homebrew/include,
+				);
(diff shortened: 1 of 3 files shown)


allejo (approved):
I don't have an M1 to test, but this seems sane to me 👍

---

Simply for our own documentation purposes,

- [Brew uses `/opt/homebrew/` for M1 Apple devices](https://docs.brew.sh/Installation)
- [Apple provides the `__aarch64__` macro for ARM devices](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary#Wrap-Platform-Specific-Code-with-Conditional-Compilation-Macros)

Xcode/BZFlag.xcodeproj/project.pbxproj
```diff
@@ -7309,9 +7309,13 @@
 					../include,
 					../Xcode,
 					/usr/local/include,
+					/opt/homebrew/include,
```
jwmelto, line 7312: Is this a change in Homebrew behavior? They used to install everything in /usr/local.

allejo: Yea, see my review. For M1 devices, they've moved installation to `/opt/homebrew/`

jwmelto: well that's unfortunately inconsistent. And it appears that Homebrew abandoned IRC (there used to be a freenode channel that was relatively active).

https://docs.brew.sh/Installation shows:
* `/usr/local` for macOS Intel
* `/opt/homebrew` for Apple Silicon
* `/home/linuxbrew/.linuxbrew` for Linux

Oh well. Looks good to me too.

jwmelto: I wonder if this should (could) be contextual? That is, use /opt/homebrew for M1 and /usr/local for Intel?
https://docs.brew.sh/FAQ#why-is-the-default-installation-prefix-opthomebrew-on-apple-silicon

jwmelto: Alternatively, `/opt/homebrew` should be before `/usr/local`, so you find native libraries first. Although I'm not sure that's an alternative. As this PR stands, things could be found in either directory, leading to confusing mixed architecture build errors. Especially with the preference for `/usr/local`.

sanssecours: Thank you both for your helpful comments. The header and library paths should now depend on the architecture. My knowledge of Xcode is very (very) limited though. At least on my machine the current solution seems to work.

macsforme:
Did this build successfully without changing the deployment target?  M1 only has Big Sur and later, and when I did my own M1 build a few weeks ago I thought I had to change the deployment target.  I believe our 2.4 builds still are set to a 10.7 (Lion) deployment target, while in master we set it to 10.9 (Mavericks), which I thought was necessary to get it to build on newer macOS versions.

sanssecours:
I also thought that I might have to change the deployment target to a newer version of macOS. However, it seems that building still works on my computer, when the macOS deployment target is set to `10.7`.

macsforme:
> I also thought that I might have to change the deployment target to a newer version of macOS. However, it seems that building still works on my computer, when the macOS deployment target is set to `10.7`.

Okay, I am building your PR on my M1 on Monterey, and I am seeing warnings for the deployment target mismatch, but at least no errors.  I am assuming homebrew doesn't specify a deployment target for the libraries it installs?

When building for an actual release, I've been building in /usr/local, and specifying the deployment target (10.7) when building each library as in our README file.  This was necessary to get everything to match up.  I'm thinking homebrew will be fine for people building for themselves (these warnings about the deployment target will manifest), but the dependencies will still need to be manually built for a release.

What about SDL2?  Did you install SDL2 from homebrew, or copy their pre-built (Intel-only, I believe) framework into /Library/Frameworks?  The build fails if /Library/Frameworks/SDL2.framework is not present, but now I'm uncertain how SDL2 is being linked in, and whether SDL2 is also the ARM version.

Overall I think this PR has the right steps.  I just want to ensure we have considered every pertinent factor.

sanssecours:
> Okay, I am building your PR on my M1 on Monterey, and I am seeing warnings for the deployment target mismatch, but at least no errors. 

Thanks for the hint. I just ignored all warnings completely 😅.

> I am assuming homebrew doesn't specify a deployment target for the libraries it installs?

Honestly, I did not think about that at all when I built BZFlag on my machine. It seems [`otool` should be able to tell use the minimum required macOS version](https://stackoverflow.com/questions/17143373/determine-minimum-osx-version-a-binary-was-compiled-for). For example, according to the output of

```sh
otool -l /opt/homebrew/lib/libGLEW.2.2.0.dylib
```

(and the warning output of Xcode) the Homebrew version of GLEW on Apple Silicon Macs seems to require macOS `12.0`:

```
Load command 10
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform 1
    minos 12.0
      sdk 12.0
   ntools 1
     tool 3
  version 711.0
```

> When building for an actual release, I've been building in /usr/local, and specifying the deployment target (10.7) when building each library as in our README file. This was necessary to get everything to match up. I'm thinking homebrew will be fine for people building for themselves (these warnings about the deployment target will manifest), but the dependencies will still need to be manually built for a release.

That sounds like the right approach 👍, since I also noticed that the version of BZFlag I built links to the Homebrew version of GLEW and c-ares:

```sh
otool -L /Applications/BZFlag.app/Contents/MacOS/BZFlag-2.4.23
# …
#	/opt/homebrew/opt/glew/lib/libGLEW.2.2.dylib (compatibility version 2.2.0, current version 2.2.0)
# …
#	/opt/homebrew/opt/c-ares/lib/libcares.2.dylib (compatibility version 2.0.0, current version 2.5.1)
# …  
```

> What about SDL2? Did you install SDL2 from homebrew, or copy their pre-built (~Intel-only, I believe~) framework into /Library/Frameworks? 

I copied the the SDL framework directory (downloaded from the official homepage) and pasted it in `/Library/Frameworks`, since Xcode was unable to locate the Homebrew version of the library.

> Overall I think this PR has the right steps. I just want to ensure we have considered every pertinent factor.

That makes sense. I usually prefer more scrutiny before something is merged (or rejected) anyway 😊. While that usually means more work, I also learn a lot more in the process. Thank you for your help and also for building/providing the official binaries for macOS.

macsforme (approved):

macsforme:
I discovered that the current per-architecture settings in the Xcode project do not add /usr/local/include and /usr/local/lib to the respective search paths when building on ARM (only /opt/brew/include and /opt/brew/lib are added).  The logic looks right in the Xcode project, so I have to do further research to see why this is.  Also, now it isn't using the /usr/local paths at all for me unless I add them under "System Header Search Paths" versus "Header Search Paths" (I'm not sure why this is either; the /usr/local paths were formerly under "Header Search Paths" just as they are in this PR).  Finally, I am looking into per-architecture deployment targets, but so far this is not working either.  It appears that the deployment target is fed into the build system in two different places, and those two places are not getting the same value.  Let me investigate these issues before we merge this.

sanssecours:
> I discovered that the current per-architecture settings in the Xcode project do not add /usr/local/include and /usr/local/lib to the respective search paths when building on ARM (only /opt/brew/include and /opt/brew/lib are added).

This behavior is actually what I was expecting when I set the values. As you noted, it does make sense to add the `/usr` paths on Apple Silicon Macs too. I changed the behavior in the last commit accordingly. Now building on my M1 Mac works, if I install `c-ares` and `glew` according to the instructions of [`README.MacOSX`](https://github.com/BZFlag-Dev/bzflag/blob/2.4/README.MacOSX). Compiling still works after I removed the Homebrew version of GLEW.

macsforme:
I don't know what the issue was with my "Header Search Paths" versus "System Header Search Paths," but it seems to have resolved itself, possibly with your last commit.

The deployment target will just have to be 10.9 on both architectures.  I was not able to determine how to do per-architecture deployment targets and have it actually work.

There is one more issue with the screen being offset when you hit F1 the first time, but I just checked and it happens on the Intel side too with later SDL2 versions, so it doesn't appear to be specific to Apple M1.

I think this is good to go, unless anyone else has further concerns.

PR #302, opened 2022-01-26
Do not use isEventPending to exit from event loop

atupone:
isEventPending is not really needed on the event loop, as there is already a check in the peekEvent for no events.
On the curl loop, the other place where  isEventPending is present, added a check for getEvent not returning any event.

With these changes, SDL_PollEvents(NULL) is never called


atupone:
It works only for SDL

PR #307, opened 2022-03-31
Fix incorrect rotation for API custom zones

allejo:
After chatting with tupone on IRC, he pointed out that the logic for determining whether a point was inside of a custom zone was incorrect when it was rotated 90 or 270 degrees. Originally when I tested this code, I made the mistake of testing it when an object at coordinates (0,0) so it worked just fine; but if the object as anywhere but the origin, it'd be incorrect. Additionally, the debug message for BBOX conversion outputted an incorrect position; that has also been fixed.

### DIFF
--- a/src/bzfs/bzfsAPI.cxx
+++ b/src/bzfs/bzfsAPI.cxx
@@ -3114,15 +3114,10 @@ BZF_API bool bz_CustomZoneObject::pointInZone(float pos[3])
 {
     if (box)
     {
-        if (rotation == 0 || rotation == 180)
+        if (fmod(rotation, 90) == 0)
         {
             if (pos[0] > xMax || pos[0] < xMin) return false;
             if (pos[1] > yMax || pos[1] < yMin) return false;
-        }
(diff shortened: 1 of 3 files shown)


src/bzfs/bzfsAPI.cxx
```diff
@@ -3272,13 +3267,24 @@ BZF_API void bz_CustomZoneObject::handleDefaultOptions(bz_CustomMapObjectInfo *d
     {
         if (box)
         {
-            xMin = _pos[0] - _size[0];
-            xMax = _pos[0] + _size[0];
-            yMin = _pos[1] - _size[1];
-            yMax = _pos[1] + _size[1];
             zMin = _pos[2];
```
atupone, line 3270: I wonder if it is not better for the whole test to translate the position to test in a coordinate sistem where 0,0 is the middle of the box or cylinder, and then, in case of the box to anti-rotate the point against the 0,0 coordinate.

The test then is very easy to do

atupone (changes requested):

atupone:
applied

PR #308, opened 2022-03-31
Rewrote check point in zone

atupone:
Simplify the calculations needed for checking if a point is inside a given custom zone.

Closes #307 as this is a more efficient calculation

### DIFF
--- a/include/bzfsAPI.h
+++ b/include/bzfsAPI.h
@@ -2000,11 +2000,20 @@ public:
     bool box;
     float xMax,xMin,yMax,yMin,zMax,zMin,radius,rotation;
 
+    // Half height and width only for boxes
+    float hh;
+    float hw;
+
+    // Center of the zone
+    float cX;
+    float cY;
+
+    // Used for rotation
+    float cos_val;
+    float sin_val;
+
     BZF_API bool pointInZone(float pos[3]);
     BZF_API void handleDefaultOptions(bz_CustomMapObjectInfo *data);
-
-private:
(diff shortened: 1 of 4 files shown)


include/bzfsAPI.h
```diff
@@ -2000,11 +2000,20 @@ class bz_CustomZoneObject
...
+    float hw;
+
+    // Center of the zone
+    float cX;
+    float cY;
+
+    // Used for rotation
+    float cos_val;
+    float sin_val;
+
```
allejo, line 2014: Do we want to make these to be public? The previous `xMax` and friends were public kinda by mistake (blame younger me). Maybe `protected` for these instead?

allejo (approved):
Tested and can confirm this is working! A much simpler and quicker implementation, I love it 🥳

include/bzfsAPI.h
atupone: Those attributes are used in a function that is not a member (the one to generate random point in a zone). I wonder why this function is not a member.

allejo: Ah you're right, I missed that fact! It's partly the legacy design for the API *and* it has to do with compiled libs. I have a pretty poor understanding of things, but I believe that changing things inside of a class (e.g. adding new functions or changing its signature) could cause plugins compiled with older versions of bzfs to crash at runtime because offsets in memory addresses could be different. But if it's compiled this way, the offsets of memory addresses should be relatively safe to access

PR #311, opened 2022-06-26
add missing bzfsAPI function bz_getPlayerCurrentState()

devindelong:
Committer: Devin Delong <devindelong@protonmail.com>

bz_getPlayerCurrentState() is documented on bzflag.org, but I could not find it in code. Return value was changed to bool.

### DIFF
--- a/include/bzfsAPI.h
+++ b/include/bzfsAPI.h
@@ -1556,6 +1556,7 @@ BZF_API bool bz_isPlayerPaused( int playerID );
 BZF_API double bz_getPausedTime( int playerID );
 BZF_API double bz_getIdleTime( int playerID );
 
+BZF_API bool bz_getPlayerCurrentState(int playerID, bz_PlayerUpdateState &state);
 BZF_API bz_eTeamType bz_getPlayerTeam(int playerID);
 BZF_API const char* bz_getPlayerCallsign(int playerID);
 BZF_API const char* bz_getPlayerMotto(int playerID);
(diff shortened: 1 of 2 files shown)


allejo (changes requested):
The way our API is designed, all objects are always created and freed by the server; e.g. `bz_getPlayerByIndex` + `bz_freePlayerRecord`, `bz_newIntList` + `bz_deleteIntList`. Right now, the only way to get a `bz_PlayerUpdateState` in a plugin is via a player record like so,

```cpp
bz_BasePlayerRecord *pr = bz_getPlayerByIndex(0);

pr->lastKnownState->velocity;

bz_freePlayerRecord(pr);
```

Since this function erroneously exists in our documentation, we have the freedom to change its signature and make it more consistent with our API design. Maybe something like,

```cpp
BZF_API bz_PlayerUpdateState* bz_getPlayerCurrentState(int playerID);
BZF_API void bz_freePlayerUpdateState(bz_PlayerUpdateState *state);
```

Not sure if that actually adds any extra benefit or convenience, besides a micro-optimization of slightly less information getting copied over versus getting a full player record.

devindelong:
The reason I was interested in this function was to avoid dynamic memory allocation. There are some other lightweight functions for querying player information, such as the team. I don't think it's necessary unless a standard container is involved.

Basically I was thinking that allocating memory inside of a PlayerUppdateEvent could be quite wasteful. I don't know how often it updates. If I assume that it updates 30 times per second for every single player, then, if there are N players, then there are 30*N allocations/deallocation per second. The player records also have pointers to standard vectors, which are also allocated, and those have allocators that also allocate/reserve memory blocks of some arbitrary size. If there are 10 players on the server, then there are 300 allocations/deallocations per second, constantly, which is just unnecessary.

I thought about adding a player manager that keeps track of current players, but there is already a player manager in the game. It's also nicer to not have to allocate memory.

Issue #313, opened 2022-08-07
Blocktime is calculated wrong in ServerLink.cxx


DeathByDenim:
I noticed that bzadmin was taking up much more CPU resources than I expected it too. It turns out that the reason for that is that `select` isn't blocking for the right amount of time due to an issue with the calculation of the `struct timeval`.

This happens both [src/bzadmin/ServerLink.cxx](https://github.com/BZFlag-Dev/bzflag/blob/29c3ab0cebd5f92e5bd47eaad5ad4bb708f32d23/src/bzadmin/ServerLink.cxx#L549-L551) as well as [src/bzflag/ServerLink.cxx](https://github.com/BZFlag-Dev/bzflag/blob/29c3ab0cebd5f92e5bd47eaad5ad4bb708f32d23/src/bzflag/ServerLink.cxx#L503-L505) in these lines:
```
    struct timeval timeout;
    timeout.tv_sec = blockTime / 1000;
    timeout.tv_usec = blockTime - 1000 * timeout.tv_sec;
```
`timeout.tv_sec` is correct since `blockTime` is in msec, but `timeout.tv_usec` is not. It gets the remaining number of milliseconds instead of the number of microseconds. So, it's off by a factor of 1000.

PR #314, opened 2022-08-07
Fix blocking time calculation

DeathByDenim:
Fix for issue #313. This corrects the calculation from block time to a `struct timeval` from milliseconds to microseconds.

### DIFF
--- a/src/bzadmin/ServerLink.cxx
+++ b/src/bzadmin/ServerLink.cxx
@@ -541,7 +541,7 @@ int ServerLink::read(uint16_t& code, uint16_t& len, void* msg, int blockTime)
     // block for specified period.  default is no blocking (polling)
     struct timeval timeout;
     timeout.tv_sec = blockTime / 1000;
-    timeout.tv_usec = blockTime - 1000 * timeout.tv_sec;
+    timeout.tv_usec = 1000 * (blockTime % 1000);
 
     // only check server
     fd_set read_set;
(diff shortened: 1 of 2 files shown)


src/bzadmin/ServerLink.cxx
```diff
@@ -541,7 +541,7 @@ int ServerLink::read(uint16_t& code, uint16_t& len, void* msg, int blockTime)
     // block for specified period.  default is no blocking (polling)
     struct timeval timeout;
     timeout.tv_sec = blockTime / 1000;
-    timeout.tv_usec = blockTime - 1000 * timeout.tv_sec;
+    timeout.tv_usec = 1000 * (blockTime % 1000);
```
jwmelto, line 544: This is more correct than what was there, presuming that `blockTime` is actually provided in milliseconds. Unfortunately, there's not enough context here to know.

DeathByDenim: It is actually in milliseconds as far as I can tell. There's a comment in [src/bzadmin/BZAdminClient.cxx:150](https://github.com/BZFlag-Dev/bzflag/blob/e55a68ce79084864dca2ff7ec8565b022a640a32/src/bzadmin/BZAdminClient.cxx#L150) referring to this.
```
// read until we have a package, or until we have waited 100 ms
if (sLink.read(code, len, inbuf, 100) == 1)
```

Additionally, the calculation `timeout.tv_sec = blockTime / 1000;` also suggest that the blockTime is milliseconds.

DeathByDenim:
Oh, I saw that this PR was applied to the 2.4 branch (ec6fff321415566b2ac404aef944fcff9b026800)

@jwmelto : Since you are the one reviewing this PR, does this mean this PR should be closed?

atupone:
I think you should leave opened. This should be applied on master together with another commit that make tcp read non blocking. Both affect the same piece of code.

DeathByDenim:
Ah ok, thanks for the clarification!

atupone:
Applied. Thanks

PR #316, opened 2022-10-20
Fix mouse wheel firing causing MG to endlessly fire

blast007:
Our mouse wheel event only triggered a KeyDown event. When mouse wheel was mapped to fire, this would cause MG to fire endlessly. This change adds a custom user event that gets pushed to SDL when a mouse wheel event is triggered. When this custom event is processed, we trigger a KeyUp event for the same mouse wheel direction.

### DIFF
--- a/src/platform/SDL2Display.cxx
+++ b/src/platform/SDL2Display.cxx
@@ -16,7 +16,7 @@
 // System includes
 #include <vector>
 
-SDLDisplay::SDLDisplay()
+SDLDisplay::SDLDisplay() : mouseWheelStopEvent(SDL_RegisterEvents(1))
 {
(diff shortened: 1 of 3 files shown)


src/platform/SDL2Display.cxx
```diff
@@ -78,6 +78,9 @@ SDLDisplay::SDLDisplay()
 
     // register modes
     initResolutions(_resolutions, _numResolutions, defaultResolutionIndex);
+
+    // Register a user event so we can trigger a KeyUp event for the mouse wheel
+    mouseWheelStopEvent = SDL_RegisterEvents(1);
```
jwmelto, line 83: can this be done in the initializer list?

macsforme (approved):

PR #317, opened 2022-11-02
Update CodeQL workflow

blast007:
This PR will update the CodeQL actions to the v2 versions since Github is deprecating the v1 actions. It will also build the game client and bzadmin with ncurses instead of just the server and a basic bzadmin.


jwmelto:
I have a love/hate relationship with cache

blast007:
My two attempts at getting the cache to work haven't worked anyway, so I'll probably just forget about that part for now.

PR #318, opened 2022-12-04
Remove non-SDL2 platform code and add controller rumble support

blast007:
This is still a work in progress. I might also remove the directional force feedback support from the game since there really isn't any hardware that supports that anymore. This still uses the normal Joystick API from SDL, not the Game Controller API. I would wait to implement that for BZFlag 2.6, along with enhancements to the input settings and controller mapping.

I've been testing this on Windows 10, Debian 11, and macOS 11.6.5 with a Playstation Dual Shock 4, 8BitDo SN30 pro wired, 8BitDo SN30 pro bluetooth, and 8BitDo SN30 pro+. The SN30 pro wired doesn't work on macOS. With the 8BitDo bluetooth devices, they work best in either XInput mode or mac mode, with the DInput mode usually (or always?) not working with rumble. Rumble also worked fine on the PS4 (with all three platforms) and 8BitDo SN30 pro wired (on Windows and Linux).


macsforme:
Sorry I didn't get around to commenting earlier on this.  I did a few cursory looks over the changes and had no concerns.  I am sure that if there are any outstanding issues they can be resolved in PR #255 (or later).

PR #321, opened 2023-02-03
use the new MIME interface for cURL

atupone:
Is not tested. Only compiled.
There is no call of that code on the tree, and is not exported.
But maybe in the future could be.


blast007 (approved):
This all seems to work fine for me.

PR #323, opened 2023-03-02
IPv6 work in progress

timriker:


blast007:
There's still remaining build errors and warnings with Visual Studio.

In Address::getIPVersion(), sin6_family is of type ADDRESS_FAMILY, which is typedef'ed to be an unsigned short with Winsock, so there's a warning about a loss of data converting it to an uint8_t. There's the same warning in Address::pack(...) for ```buf = (unsigned char*)nboPackUByte(_buf, addr.sin6_family);```.  We probably want to have our own defines for this anyway since the value for AF_INET6 on Windows is 23 but it is 10 on Linux.

Then there's a bunch of errors about __in6_u not being a member of in6_addr (which is the type of sin6_addr used in quite a number of places).  With Winsock, the in6_addr type only has a ```u``` member with further has ```unsigned char Byte[16]``` and ```unsigned short Word[8]``` members, so it doesn't look like Winsock exposes 32-bit chunks of an IPv6 address.

timriker:
I've been busy with other projects, sorry. I want to get windows builds working for me locally. If port is being cast to a uint8 someplace, that's a bug. Ports should be 16 bit.

blast007:
It was the address family being cast to uint8_t, not the port.

PR #328, opened 2023-06-14
Xcode project fixes

### LINKED ISSUE #297: Document Xcode project failing because of CodeSign
If you run into this problem when building locally

```
/Users/allejo/Library/Developer/Xcode/DerivedData/BZFlag-uuid/Build/Products/Debug/BZFlag-2.4.23.app: code object is not signed at all
In subcomponent: /Users/allejo/Library/Developer/Xcode/DerivedData/BZFlag-uuid/Build/Products/Debug/BZFlag-2.4.23.app/Contents/Frameworks/SDL2.framework
Command CodeSign failed with a nonzero exit code
```

here's _one_ way of fix it

1. Go to "Build Settings" for the main BZ target
   image
2. Select "Code Signing Identity" and choose "Other"
   image
3. Erase the contents so that it's empty and hit Enter. This'll remove the value from the "Code Signing Identity" field and let you build normally for local
   image
   image

macsforme:
This is the second half of [bzflag-dependencies/11](https://github.com/BZFlag-Dev/bzflag-dependencies/pull/11). This includes many updates to the Xcode project. From the commit message:

- Libraries and headers in the dependencies package are now prioritized if present.
- Dependency license files have been removed and are now copied from the dependencies package if present.
- Reset most compiler warnings to Xcode defaults:
  - We never compiled cleanly on Clang anyway...
  - There are many warnings about implicit conversion/precision loss, unsafe functions like sprintf, etc.
  - There are a few warnings about using deprecated functions (beyond OpenGL) which warrant our attention.
- Updated/modernized several general build settings (C++ dialect, C standard library, etc.).
- Updated, fixed, and/or eliminated numerous target-specific build settings:
  - Many target build settings (especially for plugins) were superfluous or incorrect.
  - There is now a more consistent inheritance hierarchy and better consistency in build target settings.
  - Several issues with file paths were corrected.
- CFBundleVersion was changed to 0 (this is actually an internal build number for app store submissions, etc.).
- Binaries and libraries are now ad-hoc signed (required for Apple Silicon).
- Changed the deployment target to 10.13 (the lowest supported by Xcode 14).
- Cross-compiling between x86_64 and arm64 now works.
- Various other fixes.

Fixes #297


allejo (approved):
I've gone through *most* of the setting changes in this PR and things look good to me during my reviewing as well as building on my ARM machine. I'll check later tonight to make sure Intel macOS building is working still, though I can't imagine anything breaking from this.

PR #334, opened 2023-09-30
Use glm

atupone:


atupone:
This is now obsolete. I am reworking it

PR #339, opened 2024-01-16
Add official Dockerfile to the project

allejo:
Creating this PR so that if any discussions need to happen, they can happen here. Right now, I'm just documenting how to use the Dockerfile

## Building the container

```
docker build --build-arg="configure=--enable-custom-plugins-file=''" .
```

## Running the container

```
docker run -p 5154:5154 -p 5154:5154/udp
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM alpine:3.18
 ARG BZ_VERSION
 
 RUN apk update && apk upgrade --no-cache
-RUN apk add autoconf automake c-ares-dev curl-dev g++ git glew libtool make ncurses zlib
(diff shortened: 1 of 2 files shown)


blast007:
We probably want to add some LABEL instructions to our Dockerfile.

https://docs.docker.com/engine/reference/builder/#label
https://snyk.io/blog/how-and-when-to-use-docker-labels-oci-container-annotations/
https://github.com/opencontainers/image-spec/blob/main/annotations.md
https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry

blast007:
Is there anything else to add to this before it would get merged?

allejo (commented):
I can't approve my own PR but I'm loving your changes @blast007. I think this is ready to be merged in

PR #340, opened 2024-02-18
Lars addingsplit

lars-vc:


bz-next:
Is there some way to avoid having to call .val() on protected values? This would necessitate a bunch of code changes and I worry it'd make integration a bit messy.

lars-vc:
Oh oops, I did not intend on this pr to be here. This was just for my fork :facepalm:.
But yes, I am personally not a huge fan of the .val() too, I am afraid there might be no other way, sadly :/.
I thought maybe something with the operator overloading of the cast (so like `operator float() const { return val;}`) but that causes ambiguous behaviour.

Sorry for making you look at the monstrosity of a pr. :sweat_smile: 
If you do know any smart idea to solve this val issue, do let me know :smile: !

PR #359, opened 2024-09-14
Lots of typo fixes

TheGiraffe3:
You might want to squash and merge.

Automated processes found most of these errors.


AUTHORS
TheGiraffe3, line 1: I'm not sure what happened here. I fixed the first error but did nothing else.

On other OSes, does it show as removing sensible characters and adding question marks, or vice versa?
My operating system says I fixed a whole bunch of question marks.

AUTHORS
```diff
@@ -111,7 +111,7 @@ mention and appreciation for their involvement.
 
 -- 1993 --
 Phillip Ezolt
-Bj�rn Augustsson
+Björn Augustsson
```
blast007: Your editor seems to have mishandled the character encoding of the file and broke some of the author names.

autogen.sh
```diff
@@ -988,7 +988,7 @@ initialize ( ) {
     if test ! -d "$_aux_dir" ; then
 	_aux_dir=.
     else
-	$VERBOSE_ECHO "Detected auxillary directory: $_aux_dir"
+	$VERBOSE_ECHO "Detected auxilary directory: $_aux_dir"
```
blast007: This should actually be "auxiliary"

autogen.sh
```diff
@@ -1110,7 +1110,7 @@ if [ "x$HAVE_AUTORECONF" = "xyes" ] ; then
 	$ECHO "Warning: $AUTORECONF failed"
 
 	if test -f ltmain.sh ; then
-	    $ECHO "libtoolize being run by autoreconf is not creating ltmain.sh in the auxillary directory like it should"
+	    $ECHO "libtoolize being run by autoreconf is not creating ltmain.sh in the auxilary directory like it should"
```
blast007: This should actually be "auxiliary"

m4/sdl2.m4
```diff
@@ -196,7 +196,7 @@ int main(int argc, char *argv[])
           echo "*** If you have an old version installed, it is best to remove it, although"
           echo "*** you may also be able to get things to work by modifying LD_LIBRARY_PATH"],
         [ echo "*** The test program failed to compile or link. See the file config.log for the"
-          echo "*** exact error that occured. This usually means SDL was incorrectly installed"
+          echo "*** exact error that occurred. This usually means SDL was incorrectly installed"
```
blast007, line 199: This typo is from upstream SDL repository that we pulled this file from.  Fixing it upstream would be better, and then we can fix it here by pulling it from upstream.

https://github.com/libsdl-org/SDL/blob/SDL2/sdl2.m4

misc/libtool.m4
```diff
@@ -1033,7 +1033,7 @@ if test -n "$_LT_AC_TAGVAR(hardcode_libdir_flag_spec, $1)" || \
    test -n "$_LT_AC_TAGVAR(runpath_var, $1)" || \
    test "X$_LT_AC_TAGVAR(hardcode_automatic, $1)" = "Xyes" ; then
 
-  # We can hardcode non-existant directories.
+  # We can hardcode non-existent directories.
```
blast007: The typos in this file are also from an upstream project, but they've since fixed it.  I'm actually not sure why we have misc/libtool.m4 when we also have m4/libtool.m4 (which doesn't have this typo).  Maybe this copy of the file could just go away.

misc/shtool
```diff
@@ -267,7 +267,7 @@ eval `echo h.$opt_spec |\
 eval `echo h:help,$opt_alias |\
       sed -e 's/-/_/g' -e 's/\([a-zA-Z0-9]\):\([^,]*\),*/opt_ALIAS_\2=\1;/g'`
 
-#   interate over argument line
+#   iterate over argument line
```
blast007, line 270: This is a generated file from an upstream project called [GNU shtool](https://www.gnu.org/software/shtool/shtool.html). We're using an older version, but the latest release still has these typos. Given that the newest release is from 2008, I don't imagine this will get fixed upstream.

data/l10n/bzflag_sv.po
```diff
@@ -1276,10 +1276,10 @@ msgstr "högt uppe den är: Ju större prick, desto högre upp. Flaggor är små
 #msgid "Toggle Main Flags:"
 #msgstr ""
 
-#msgid "Silience/UnSilence"
+#msgid "Silence/Unsilence"
```
blast007, line 1279: There are more instances of this typo in the other language files. The src/bzflag/HelpMenu.cxx and src/bzflag/KeyboardMapMenu.cxx have the S in UnSilence capitalized. That's unnecessary there as well. The data/l10n/bzflag_lt.po has a typo on the line for Silence/UnSilence that reads "# sgid" instead of "#msgid".

blast007 (changes requested):
Most of this seems fine.  The AUTHORS file needs the encoding fixed so it doesn't break names.

For any msgid changes in the data/l10n/ files, you'd want to sure that the strings in the game code are changed as well.  For instance, you changed "recieved" to "received", but src/bzadmin/ServerLink.cxx and src/bzflag/ServerLink.cxx still have "bytes recieved".

I'm not sure if we should care about fixing typos shtool since it *is* a third-party file and it isn't one that anyone every looks at anyway.

I'd need to check a bit more if we can just get rid of misc/libtool.m4.

AUTHORS
TheGiraffe3: Interesting. On my operating system it says it had fixed them.

Did e03c3a0 make this for you? It just got worse on my computer.

autogen.sh
TheGiraffe3:
```suggestion
	$VERBOSE_ECHO "Detected auxiliary directory: $_aux_dir"
```

autogen.sh
TheGiraffe3:
```suggestion
	    $ECHO "libtoolize being run by autoreconf is not creating ltmain.sh in the auxiliary directory like it should"
```

misc/shtool
TheGiraffe3: So would you like it to be fixed here or stay consistent with the other release?

m4/sdl2.m4
TheGiraffe3: Changes suggested and merged. I'm not sure how to pull from upstream, so should I just leave it like this?

include/Obstacle.h
```diff
@@ -59,7 +59,7 @@ class Obstacle
         @param hwidth     Half the X size of the obstacle
         @param hbreadth   Half the Y size of the obstacle
         @param height     The Z size of the obstacle
-        @param drive      @c true if the obstacle is drivethtrough, i.e. tanks
+        @param drive      @c true if the obstacle is driveththrough, i.e. tanks
```
TheGiraffe3:
```suggestion
        @param drive      @c true if the obstacle is drivethrough, i.e. tanks
```

include/Obstacle.h
```diff
@@ -38,7 +38,7 @@ class MeshTransform;
 
 /** This ABC represents a (normally) solid object in a world. It has pure
     virtual functions for getting information about it's size, checking ray
-    intersections, checking point intersections, computing normals etc.
+    intersections, checking point intersections, computting normals etc.
```
TheGiraffe3:
```suggestion
    intersections, checking point intersections, computing normals etc.
```

TheGiraffe3 (commented):
Not sure what happened here.

TheGiraffe3:
@blast007 is there something in particular this PR is waiting on?
Would you like me to close and open a new one that only has one commit but all these changes?

TheGiraffe3:
Will reopen as a new PR later.

EDIT: #362.

PR #367, opened 2024-11-28
Kick all replay viewers when an incompatible replay is loaded

### LINKED ISSUE #348: Crash on 2.4.27 corrupted size vs. prev_size
I switched to 2.4.27 recently to try to help debug the next release.

I visited a replay server and poked around for a bit, then left, and tried joining a different replay server.

The client crashed with:

```
corrupted size vs. prev_size
Aborted (core dumped)
```

Version string: `BZFlag client 2.4.27.20240320-DEVEL-linux-gnu-SDL2 (protocol 0221) http://BZFlag.org/`
uname -a `Linux fedora 6.7.9-200.fc39.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Mar  6 19:35:04 UTC 2024 x86_64 GNU/Linux`

I didn't actually have core dumps enabled on my OS, so I don't have a backtrace.

I'll update here if I can reproduce it.

blast007:
Rejoining was already required, since playing an incompatible replay without rejoining would usually crash the client.

This resolves #348

### DIFF
--- a/src/bzfs/RecordReplay.cxx
+++ b/src/bzfs/RecordReplay.cxx
@@ -153,7 +153,7 @@ static bool savePacket(RRpacket *p, FILE *f);
 static RRpacket *loadPacket(FILE *f);       // makes a new packet
 
 static bool saveHeader(int playerIndex, RRtime filetime, FILE *f);
-static bool loadHeader(ReplayHeader *h, FILE *f);
+static bool loadHeader(ReplayHeader *h, FILE *f, bool *incompatible);
 static bool saveFileTime(RRtime filetime, FILE *f);
 static bool loadFileTime(RRtime *filetime, FILE *f);
 static bool replaceFlagTypes(ReplayHeader *h);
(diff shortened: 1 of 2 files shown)


src/bzfs/RecordReplay.cxx
```diff
@@ -832,6 +833,71 @@ bool Replay::loadFile(int playerIndex, const char *filename)
...
+         *    record file is loaded (with an appropriate warning).
+         *    then they can reload with the original DB upon rejoining
+         *    (DB with modified maxPlayers).
+         *
+         * 3) make fixed sized arrays on the client side
+         *    (but what if someone really needs 1000 flags?)
+         *
+         * 4) implement a world reload feature on the client side,
+         *    so that if the server sends a MsgGetWorld to the client
+         *    when it isn't expecting one, it reaquires and regenerates
```
TheGiraffe3:
```suggestion
         *    when it isn't expecting one, it reacquires and regenerates
```

blast007: Thanks, fixed.

TheGiraffe3 (approved):

Issue #372, opened 2025-04-22
Migrate from GLEW to libepoxy or GLAD


cryptomilk:
Hi,

we are currently trying to get rid of GLEW in Fedora as it isn't [maintained](https://github.com/nigels-com/glew/issues/416#issuecomment-2646570907) anymore. Also GLEW with the last release only supports GLX and not EGL.

There are two ways moving forward:
* Use [libepoxy](https://github.com/anholt/libepoxy)
* Use [GLAD](https://github.com/Dav1dde/glad/), or check out the [webservice for glad2](https://glad.sh/) to generate the files you need!

With both of them migration should be straight forward.

Thanks for considering!

blast007:
I'd say skip looking at libepoxy since that has annoying build dependencies.

blast007:
> GLEW with the last release only supports GLX and not EGL.

I see a build option for EGL in the GLEW CMake project, and openSUSE has a [patch](https://build.opensuse.org/projects/openSUSE:Factory/packages/glew/files/glew-rename-EGL-library.patch?expand=1) to that so it adds an _EGL suffix when building with that option.  They then have a glew:egl package for the EGL version.  Can Fedora do something similar?

I also don't really see evidence that GLEW is not maintained.

If we wanted to move to GLAD, it might require packaging the glad generator in other Linux distributions since Debian doesn't like generated or third-party code bundled in the source.  Debian does have the python3-glad package and Fedora has python3-glad2, but Gentoo doesn't seem to have it packaged in the main repository. There's probably others that don't have it packaged either.

cryptomilk:
I backported those patches in openSUSE and this is an ugly workaround.

GLAD is a code generator. python3-glad is the commandline client to generate loader code. It does the same as the web interface linked above. It is not a shared library. If you want to use a shared library then go with libepoxy. This is like libglew but doesn't have a dependency on GLX.

PR #373, opened 2025-04-24
Use GLAD2 instead of GLEW

atupone:


atupone:
Should fix https://github.com/BZFlag-Dev/bzflag/issues/372

include/bzfgl.h
```diff
@@ -17,12 +17,10 @@
...
-# error OpenGL version 1.1 functionality is required
-#endif
-
+/* GLAD2 include */
+#include <glad/gl.h>
+/* We use glu too */
+#include <GL/glu.h>
 
 /* These will track glBegin/End pairs to make sure that they match */
 #ifdef DEBUG
```
blast007: This block that defines glBegin and glEnd with extra validation fails to build now because GLAD already creates definitions for these. I don't know if there's a good way to retain this validation.

atupone: I did not build debug, so I did not catch that piece of code.
To do extra validation we could use what is written in https://github.com/Dav1dde/glad/wiki/C#Debugging

atupone:
No need to go glad now

PR #374, opened 2025-05-01
Fix joystick scaling

RighthandSon:
Adjust the joystick scaling to properly have full range of speed between jsRangeMin and jsRangeMax.
The way it worked before, the speed of the tank would be limited to the percentage of jsRangeMax.
I also added a box to the joystick testing menu so you can clearly see where the outer limits are.
image

### DIFF
--- a/src/bzflag/HUDuiJSTestLabel.cxx
+++ b/src/bzflag/HUDuiJSTestLabel.cxx
@@ -67,13 +67,13 @@ void            HUDuiJSTestLabel::doRender()
     glVertex2f(getX() + (1.0f + rangeLimit) / 2.0f * width, getY() + (1.0f + rangeLimit) / 2.0f * height);
     glEnd();
 
-    //draw a box around the test area
+    // draw a box around the full -1, 1 range
     glColor4fv(boxColor);
     glBegin(GL_LINE_LOOP);
-    glVertex2f(getX() + (1.0f - 1) / 2.0f * width, getY() + (1.0f + 1) / 2.0f * height);
-    glVertex2f(getX() + (1.0f - 1) / 2.0f * width, getY() + (1.0f - 1) / 2.0f * height);
-    glVertex2f(getX() + (1.0f + 1) / 2.0f * width, getY() + (1.0f - 1) / 2.0f * height);
-    glVertex2f(getX() + (1.0f + 1) / 2.0f * width, getY() + (1.0f + 1) / 2.0f * height);
+    glVertex2f(getX(), getY() + height);
+    glVertex2f(getX(), getY());
+    glVertex2f(getX() + width, getY());
+    glVertex2f(getX() + width, getY() + height);
     glEnd();
 
     // draw the real cursor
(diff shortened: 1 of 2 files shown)


macsforme:
Confirmed the issue.  First of all, thank you, as this is obviously a critical bug.

This fix seems to correct the joystick input in practice.  However, I did notice that in the "Test Joystick Range" menu, the "modified cursor" position is no longer correct.  The "real cursor" and modified cursor positions are supposed to roughly match up at the edges of the black box, and the modified cursor position is supposed to be clamped to the edge of the same black box (see the behavior of the stock client).  So actually, the test menu is showing the correct position currently, but the value going into the driving input is wrong.

In the section starting at at playing.cxx:955 (specifically see line 960), that is where the current coordinates are supposed to be scaled back up to the range of `jsRangeMax`.  Clearly, something is not working.  I will try to look at the math again with fresh eyes when I get a chance.

I think the outer box showing the maximum range is a great addition as well.

**EDIT:** In HUDuiJSTestLabel.cxx, if I multiply `jsx` and `jsy` by `rangeLimit`, the modified cursor position is once again correct.  I think we are on the right track, but I do want to look at this again with fresh eyes.  I vaguely recall that the section "proportionally scale the coordinates back to the range limit" was significant.

RighthandSon:
Yeah.. I noticed that, and spent a bit of time trying to work out what the best solution was. Ended up creating the PR to get more input.
Its not clear, but if you look at the screenshot I posted, you can see that the modified cursor position is in the same relative position in the full box, as the real cursor position is inside the black box. Once the real cursor is outside the black box the modified cursor is at the max.
As I'm writing this the idea came to me, suppose we clamp the real cursor position to be shown only inside the black box?

macsforme:
I do see that; however, the intent is for the black box to show the effective range, and for the gray diamond to show the effective position within that range.  As I mentioned in the edits to my original comment, multiplying the coordinates by the maximum range fixes the effective position in the test menu (including clamping to the edges of the effective range).

So, with that addition, I believe this fix/PR is correct.  I just want to trace over the logic a few more times, check various combinations of options, identify any other possible refactoring in light of this change (such as removing the divide by 1), etc.

macsforme:
I traced back over the original logic and while it was challenging to follow, the purpose of each component eventually became clear.  However, I did some refactoring for clarity and reordered the modifiers to what made more sense (to me): invert axes → stretch corners → dead zone and overall sensitivity → exponential ramp → clamp to range.  I also addressed a few quirks, such as the scaled radial dead zone formula behaving undesirably when the radius exceeded 1.  And with the fix in this PR, the range limit (overall sensitivity) actually works.

I need to validate these changes (including play testing), and then I will add them to this PR and this should be good from my perspective.

macsforme:
Apologies for the delay.  Mentally acute spare time has been scarce lately.

I just committed my refactor and cleanup to this PR.  It builds on the primary fix in this PR, rearranges the order of joystick modifiers, and does some additional cleanup.  I tested various combinations of settings with my Xbox controller, but further testing would certainly be welcome @RighthandSon.

For reference, [here](https://web.archive.org/web/20141025190105/https://www.third-helix.com/2013/04/12/doing-thumbstick-dead-zones-right.html) is the info on scaled radial dead zones.  It gets a little strange because we have a square zone rather than radial, so basically our logic (in the refactor) will only use this kind of scaling up until a radius of 1 (its primary purpose is to facilitate a dead zone and smooth scaling immediately thereafter).

With that, this PR looks good to me.  Many thanks once again for catching the original issue.

macsforme (approved):

RighthandSon:
Tested with my sidewinder 2 joystick. It works for me as well. Looks good, thank you!

PR #384, opened 2026-04-27
Recalculate team scores in FFA when a player leaves

TheGiraffe3:
On maps like Badgerking where there are a lot of deaths, you can sometimes end up with a team score of -50 while the only player on that team has a score of -2. This pull request fixes that by recalculating team scores whenever a player leaves.

### DIFF
--- a/src/bzfs/bzfs.cxx
+++ b/src/bzfs/bzfs.cxx
@@ -2961,6 +2961,15 @@ void removePlayer(int playerIndex, const char *reason, bool notify)
         int teamNum = int(playerData->player.getTeam());
         --team[teamNum].team.size;
 
+        // change team wins/losses
+        if(clOptions->gameType == TeamFFA && Team::isColorTeam(playerData->player.getTeam()))
+        {
+            int newTeamWins = team[teamNum].team.getWins() - playerData->score.getWins();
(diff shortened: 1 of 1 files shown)


Zehra:
~~I'd probably just add a check to see if it's not an observer.~~ Edit: On second thought, probably way better to check if they're red, green, purple or blue.

Plugin API event example below:
```cpp
    bz_PlayerJoinPartEventData_V1* partData = (bz_PlayerJoinPartEventData_V1*)eventData;
    if ((bz_getGameType() == eFFAGame) && (partData->record->team != eObservers)) {
      bz_setTeamWins(partData->record->team, bz_getTeamWins(partData->record->team) - (partData->record->wins));
      bz_setTeamLosses(partData->record->team, bz_getTeamLosses(partData->record->team) - (partData->record->losses));
    }
```

Team switching might need to be figured out in the long run.

TheGiraffe3:
What do you mean by "team switching might need to be figured out"?

Zehra:
> What do you mean by "team switching might need to be figured out"?

I haven't checked into it, but for the next major release (2.6), there's a feature of team switching. It probably doesn't trigger a part/leave event, which means team scores might be impacted there.

PR #387, opened 2026-07-14
Fix duplicate "Team" shown to user when auto-assigning teams

TheGiraffe3:
Before:
Screenshot 2026-07-14 at 2 44 39 PM

After:
Screenshot 2026-07-14 at 2 44 08 PM


blast007:
Thanks! This has been merged.

PR #388, opened 2026-08-10
Tetra bug

atupone:
normals is a float [4][3][3] so normals[1] is [3][3].
tmpNormals as a temporary should also be [3][3] while now is [4][3]
It is moving more things than it should probably overwriting part of the normal[3]

I also prefer to rewrite the entire dance of byte with std::swap


atupone:
Merged

PR #389, opened 2026-08-24
compareObstacles simplified

atupone:
Avoiding use of Custom RTTI using an added virtual getSortPriority to obstacle (Normal/MeshFace/MeshObstacle)
Dropped an if cascade with a simple diff.

Code looks better IMHO


atupone:
Merged with changes

PR #394, opened 2026-08-31
Use OCI-standard Containerfile

tainn:
The Open Container Initiative (OCI) adopts the open standard `Containerfile` instead of `Dockerfile` to reflect that the format is a vendor-neutral industry standard. Both use the exact same syntax, instructions and format to build container images. We should stick to the standard. This pull request renames `Dockerfile` to `Containerfile`, keeping the content the same.

Tools like `podman` and `buildah` default to `Containerfile` and fall back to `Dockerfile`. `docker` requires a flag (such as `docker build -f Containerfile`). To maintain interoperability for anyone relying strictly on `docker` defaults, I have also added a `Dockerfile` symbolic link to `Containerfile`.


blast007:
If the other two will look for Dockerfile by default anyway, I'm not seeing any benefit to renaming this file. It also doesn't look like Docker intends to support Containerfile.

tainn:
Functionally, there is no benefit. However, standards should be followed whenever possible when it comes to containerization, unless there is a very good reason not to.

blast007:
Can you point me at the OCI specification that talks about this?

timriker:
Is that this?
https://opencontainers.org/posts/blog/2025-11-04-oci-runtime-spec-v1-3/

tainn:
Ok, wow, this is embarrassing. I guess it goes to show that there is nothing you can be that sure of that you shouldn't confirm before contributing.

For the past few years, I was absolutely sure that Containerfile is an OCI-standard, vendor-neutral naming. It seems it is not. This was always a myth... and I was so certain, too... probably told a colleague or two this misinformation in the meantime, as well...

All things considered, I am glad I got this cleared up. I apologize for cluttering the requests here. I would personally still prefer Containerfile over Dockerfile just because of the vendor-neutral naming, but it seems there are zero standards or conventions backing this up, so it practically holds no true weight in my eyes anymore. Do as you will.

blast007:
Not a problem at all. It looks like it might be a push by podman (Red Hat) to use Containerfile instead of Dockerfile. The opencontainer repositories are still using Dockerfile, though I don't see them reference a specific name in any documentation.

